PR #31, opened 2017-05-12
fix jq error when cidr is a Ref

jaymccon:
when CidrIp is an object/array (a "Ref" for example ) jq would fail with:
```
jq: error (at test.template:23): endswith() requires string inputs
```
added an additional check to see if it was a string, if not return empty.

tested against this snippet:
```json
{
    "Resources": {
        "CfnNagFail" : {
          "Type" : "AWS::EC2::SecurityGroup",
          "Properties": {
                "SecurityGroupIngress": [
                    {
                        "CidrIp": { "Ref": "TestRef" } 
                    }
                ]
            }
        }
    }
}
```

### DIFF
--- a/lib/json_rules/cidr_rules.rb
+++ b/lib/json_rules/cidr_rules.rb
@@ -46,8 +46,15 @@ non_32_cidr_jq_expression = <<END
  else (
         if (.Properties.SecurityGroupIngress|type == "array")
         then (
-               select(.Properties.SecurityGroupIngress[]|select(.CidrIp|endswith("/32")|not))
-             )
+               select(.Properties.SecurityGroupIngress[]|
+                      if (.CidrIp|type == "string")
+                      then (
+                          select(.CidrIp|endswith("/32")|not)
+                      )
+                      else empty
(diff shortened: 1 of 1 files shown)


lib/json_rules/cidr_rules.rb
```diff
@@ -46,8 +46,15 @@
  else (
         if (.Properties.SecurityGroupIngress|type == "array")
         then (
-               select(.Properties.SecurityGroupIngress[]|select(.CidrIp|endswith("/32")|not))
-             )
+               select(.Properties.SecurityGroupIngress[]|
+                      if (.CidrIp|type == "string")
+                      then (
+                          select(.CidrIp|endswith("/32")|not)
```
erickascic: i think the object bit above this needs this too.  thanks for the fix!  I will incorporate with a fresh test shortly

jaymccon:
got addressed in https://github.com/stelligent/cfn_nag/commit/3cffd60cde3bd88150a08835622fbe0a9c6bcb47

PR #33, opened 2017-06-26
Checks for empty rule directory before trying to eval rules

akuma12:
This makes it so if a custom rule directory is specified, but no rules currently exist in the directory, it won't crash. This specifically helps the cfn_nag_server project, where a custom_rules directory is created, but by default is empty.

Also adds .gitignore file to ignore files generated from testing.

### DIFF
--- a/lib/cfn_nag.rb
+++ b/lib/cfn_nag.rb
@@ -206,9 +206,11 @@ class CfnNag
     rule_directories.each do |rule_directory|
       rules = Dir[File.join(rule_directory, '*.rb')].sort
 
-      rules.each do |rule_file|
-        @input_json = input_json
-        eval IO.read(rule_file)
+      if rules.length > 0
+        rules.each do |rule_file|
+          @input_json = input_json
+          eval IO.read(rule_file)
+        end
       end
     end
   end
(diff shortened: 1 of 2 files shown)


lib/cfn_nag.rb
```diff
@@ -206,9 +206,11 @@ def generic_json_rules(input_json, rule_directories)
     rule_directories.each do |rule_directory|
       rules = Dir[File.join(rule_directory, '*.rb')].sort
 
-      rules.each do |rule_file|
-        @input_json = input_json
-        eval IO.read(rule_file)
+      if rules.length > 0
```
erickascic: is this guard necessary?  if rules.length == 0 doesn't it fall through anyways?

akuma12: If a rule_directory is specified, but that directory doesn't contain any .rb files that can be eval'd, it blows up with an error.

akuma12: You know what, you're right. I think my brain wasn't caffeinated enough when I did this. I'll look deeper into what's causing that error.

akuma12:
The rules change was unnecessary. Reverting that in my fork and submitting a new pull request.

PR #46, opened 2017-10-13
Add initial Dockerfile

steynovich:
For development/run purposes it would be useful to have cfn_nag in a Docker container. This is the first step, automated build instructions can be added later.


erickascic:
added an issue #162 to push a docker image in tune with the latest releases of cfn_nag

PR #92, opened 2017-12-12
Expose a Lambda nag over HTTP without CodeCommit hooks

craigw:
This is mostly a reworking of the code in the existing Lambda example, with a few tiny bug fixes. It shows how to expose CfnNag over HTTP when wrapped in a Lambda.

### DIFF
--- a/http-lambda/.gitignore
+++ b/http-lambda/.gitignore
@@ -1,0 +1,14 @@
+*.class
+target
+/bin/
+/.settings/
+.project
+.classpath
+
+# Package Files
+*.jar
+*.war
+*.ear
+
+# Serverless directories
+.serverless
--- a/http-lambda/Gemfile
+++ b/http-lambda/Gemfile
@@ -1,0 +1,3 @@
+source 'https://rubygems.org'
+
+gem 'cfn-nag'
--- a/http-lambda/Gemfile.lock
+++ b/http-lambda/Gemfile.lock
@@ -1,0 +1,55 @@
+GEM
+  remote: https://rubygems.org/
+  specs:
+    cfn-model (0.1.12)
+      kwalify (= 0.7.2)
+    cfn-nag (0.3.15)
+      cfn-model (= 0.1.12)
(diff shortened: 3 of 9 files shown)


venkatkotta:
Hi @craigw , i tried your changes to scan cloudformation template on my local machine but seems like it is not working properly every time i run #Use the lambda it throws a error , even though my cloudformation template is fine and scan works fine without lambda
{
  "failure_count": 1,
  "violations": [
    "FATAL FAIL Illegal cfn - no Resources "
  ]

roberto-andrade:
Hello,

Could you please help me out on the following scenario:

I have a static web form that calls my AWS API, which triggers one lambda that runs the 'cfn_nag.audit' from the 'api_gateway_invoker.rb'.
I am passing the template file (.yaml) as a string in the JSON body using HTTP POST method.
The request JSON would be like this:
image

Here is the result (error):
image

- Am I missing some parameters in the request ?
- Does the file format would be byteArray or plain text ?
- Any other thoughts ?

Many Thanks !

erickascic:
is there an earlier exception that wipes out the audit_result?  what does audit_result look like (this is actually cfn-nag-pipeline code vs. cfn-nag itself?)

roberto-andrade:
Hi Erick,

Thanks for your support.

This code is cfn-nag-pipeline calling the 'api_gateway_invoker' from the Java Wrapper.

Here is the full exception log:
image

Many Thanks !

erickascic:
I need some time to track down what changed/when, but I'm guessing that if on line 22 of api_gateway_invoker, you can change audit_result['failure_count'] to audit_result[:failure_count] to get past the failure.  the keys in audit_result are symbols not strings, so audit_result['failure_count'] will yield a nil and then nil > 0 raises the exception.

erickascic:
Given that Lambda supports Ruby natively now, going to close this one out.  We'll have some work coming out soon to expose cfn_nag as a service but it won't leverage JRuby/java8

PR #94, opened 2017-12-18
update the pom.xml

mendickxiao:
Build failed for the initial version, there isn't jruby repos in the maven, it should be jruby-complete.


erickascic:
Are you using the great maven repo in the sky?  I don't have any trouble building and I see jruby here:

http://mvnrepository.com/artifact/org.jruby/jruby/9.1.13.0

PR #97, opened 2017-12-22
Basic documentation updates

drmmarsunited:
Added sublist steps to rule creation crash course to call out camelcase properties of unmodeled objects

Added clearer representation of the --rule-directory switch for cfn_nag_scan


drmmarsunited:
Decided to make a few more changes.  Will resubmit in a few.

PR #104, opened 2018-02-12
Don't trigger warning when pure IAM role in-use

teo-scout24:
The existing rule is a bit too sensitive. Whenever it sees `AWS::CloudFormation::Authentication` resource then warning is triggered. 

However, sometimes there is no credential when pure IAM role in use hence we should only trigger warning when accessKeyId, secretKey or password is found.


teo-scout24:
Hi @erickascic Would you be kind to merge this PR please? This has been tested to work.

teo-scout24:
Hi @erickascic  @jeffb4 @jeffb-stell  Can you please look into this PR and merge it if there isn't any issue? Thanks!

erickascic:
I'm pulling this in, but have to make two changes (so stealing the code from PR and just wrapping it up in a commit to master):

1. Any PR unless it's documentation or trivial needs a spec to go along with it to demonstrate the bug it's fixing or the feature it's adding, so I am adding a spec to go along...
2. instead of adding the authentication identifier, going to add the parent resource id for the violation

erickascic:
btw thanks for your patience on this one.

Issue #48, opened 2017-10-18
Add rule to look for RDS publicly accessible


erickascic:
Given an AWS::RDS::Instance with logical resource 'foo'
  And PubliclyAccessible is true
When cfn-nag runs
 Then 'foo' is flagged with a violation

If somebody has a burning reason to make it public, then I'm thinking they can make a profile that doesn't include the rule

erickascic:
beware of String vs. boolean

erickascic:
Fixed in 0.3.22

PR #109, opened 2018-03-15
Custom rule to check publicly accessible property in RDS instance

xysr89:
https://github.com/stelligent/cfn_nag/issues/48


erickascic:
Thanks for the PR!  Took it almost whole cloth - I added a "negative" test to make sure it doesn't flag a violation when PubliclyAccessible != true and then I changed to logic a little to flag when PubliclyAccessible isn't explicitly set.  The implicit value can vary depending on context, so want to force the developer to be aware and make a decision

xysr89:
Thanks for the clarification! I was of the understanding that for some conditional properties, we dont need to check them explicitly.(especially in EC2 and RDS) I guess it is a false assumption then :smile:

PR #128, opened 2018-04-25
Fixed operand based on rule intent

### LINKED ISSUE #127: Lambda Permission Invoke Function Incorrectly Warns
The LambdaPermissionInvokeFunctionActionRule incorrectly warns when there is a lambda invokeFunction action applied to a resource of type AWS::Lambda::Permission.

Affected line:
https://github.com/stelligent/cfn_nag/blob/8a7b28408710ca9bbfe21d20b064445883e26f2d/lib/cfn-nag/custom_rules/LambdaPermissionInvokeFunctionActionRule.rb#L19

This was changed from a != to an == in commit https://github.com/stelligent/cfn_nag/commit/36031fb95b3c2de3760017039209b98c4c81f8a0

The != was actually the correct operator based on the warning text.

I believe the intent of this rule is that you should *not* have other permissions other than invokeFunction.

JamesrenaudMitel:
Should ensure permissions other than invokeFunction are not present as opposed to warning if invokeFunction is present.

Resolves #127

### DIFF
--- a/lib/cfn-nag/custom_rules/LambdaPermissionInvokeFunctionActionRule.rb
+++ b/lib/cfn-nag/custom_rules/LambdaPermissionInvokeFunctionActionRule.rb
@@ -16,7 +16,7 @@ class LambdaPermissionInvokeFunctionActionRule < BaseRule
 
   def audit_impl(cfn_model)
     violating_lambdas = cfn_model.resources_by_type('AWS::Lambda::Permission').reject do |lambda_permission|
-      lambda_permission.action == 'lambda:InvokeFunction'
+      lambda_permission.action != 'lambda:InvokeFunction'
     end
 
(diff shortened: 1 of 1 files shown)


erickascic:
indeed.  looks like some linting madness changed select to reject and messed up the rule.... and not enough testing to catch it.  I definitely do not like the reject code so will likely just back that out.  thanks for catching!

erickascic:
actually.... added a test to confirm the bad behavior, but i think it's working correctly.  the reject method is throwing out any lambda permissions that are InvokeFunction and then considering the leftovers violating.  So.... I'm going to push that test.... but I'm also going to remove the reject code.  it may be more concise but i think selecting the violators reads better than eliminating the conforming resources and considering the leftovers as violators.

PR #139, opened 2018-09-06
Add option to ignore rules globally.

mattdjones:
I wanted to be able to suppress custom rules that aren't relevant in the environment they are being deployed to without modifying a large amount of cloud formation templates.

* Modified custom_rule_loader.rb to look for a file that contains rule Ids to ignore globally
* Added template ignore.yml in the custom_rules folder
* Updated the gemspec to copy yaml files
* Update the ReadMe to detail the option to use the ignore.yml file

### DIFF
--- a/README.md
+++ b/README.md
@@ -133,6 +133,20 @@ Failures count: 0
 Warnings count: 0
 ```
 
+Alternatively, if you have existing Cloudformation templates for which you would like to suppress warnings. 
+Edit ignore.yml in the custom_rule folder.
+Add the Ids of any rules you want to globally suppress to this file
+
+Eg:
+
+```yaml
+RulesToSuppress:
+  - id: F18
+    reason: We allow this policy
+  - id: F1000
+    reason: This rule is not valid for the environment we are deploying to.
+```
+
 # Development
 
 ## New Rules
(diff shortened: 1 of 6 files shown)


mattdjones:
@erickascic

erickascic:
looks good man, thanks for the PR.  promise to get this merged in this weekend

erickascic:
do you have any automated tests to go long for the ride?  i can write them as part of the vetting process but if you have them.... will make things go quicker

mattdjones:
hey, I don't have any tests as yet, I'll give it a shot 👍

mattdjones:
Actually, I'm getting a bit stuck on getting the spec tests to run on my Wiindows set-up.
If you have time to do the tests that would be awesome. Or any advice on running the tests with VS Code.

mattdjones:
hey @erickascic  managed to spend a bit more time on this, I got the specs running and have fixed up a couple of tests that weren't working. I'll try and get a test for the new code.

erickascic:
@mattdjones sorry for the major delay, but finally coming back to take a look at this.  the feature here is meant to globally ignore particular rules?  you can _sort of_ do this already with profiles, but a profile is a whitelist versus being a blacklist.  to do something globally, we don't need to iterate over the actual resources - we can just strip out violations from rules we don't care about at the top-level.

erickascic:
0.3.74 contains this functionality.  In cfn_nag and cfn_nag_scan you can specify --blacklist-path to point to a file with the format described aboved (yaml file with RulesToSuppress which is an array of {id, reason}

mattdjones:
@erickascic  awesome!  thanks, the blacklist will be really useful. is 0.3.74 out now?

erickascic:
@mattdjones yep it's up in rubygems:

https://rubygems.org/gems/cfn-nag/versions/0.3.74

PR #151, opened 2019-01-25
Don't evaluate potentially long debug log string

### LINKED ISSUE #150: Debug log is a memory hog even when not in debug mode
I see that cfn nag performance decreases for large templates and may even accumulate GBs of memory due to this line:
https://github.com/stelligent/cfn_nag/blob/19c233ae8b266f008540118ec72e659619797e8f/lib/cfn-nag/custom_rule_loader.rb#L42

This is because the line is not lazy evaluated, causing a very long string to be created.
I've seen that commenting just this line gives a dramatic performance gain.
I'm submitting a PR to fix this.

jeshan:
Fixes #150

### DIFF
--- a/lib/cfn-nag/custom_rule_loader.rb
+++ b/lib/cfn-nag/custom_rule_loader.rb
@@ -39,7 +39,9 @@ class CustomRuleLoader
   end
 
   def execute_custom_rules(cfn_model)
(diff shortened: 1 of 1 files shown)


jeshan:
Note that I'm not a Ruby programmer and I'm not aware of the most elegant way to handle this

PR #161, opened 2019-04-09
Re-enable rubocop

jesseadams:
Goals:

* Configure a starting point for rubocop. It isn't perfect, but at least gives us a place to start
* Fix as many violations as possible
* Ensure tests still pass
* Fail Circle builds if rubocop does not pass.

As a side effect we are dropping ruby 2.2.x (no longer supported) for 2.4.x (and also 2.5.x)

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -2,7 +2,7 @@ version: 2
 jobs:
   test:
     docker:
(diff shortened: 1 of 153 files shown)


jesseadams:
Moved to 2.5.5 per our convo.

PR #168, opened 2019-04-11
Adding dockerfile for future DockerHub publishing

jesseadams:
Pre-requisite of #162

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,8 @@
+FROM ruby:2.5
+
+MAINTAINER eric.kascic@stelligent.com
+
+RUN gem install cfn-nag
+
+ENTRYPOINT ["cfn_nag"]
+CMD ["--help"]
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -0,0 +1,8 @@
+FROM ruby:2.5
+
+MAINTAINER eric.kascic@stelligent.com
+
+RUN gem install cfn-nag
```
erickascic: i'd suggest we pin this based upon the version the pipeline pushes to rubygems.... and then the label of the image can reflect that version

Dockerfile
```diff
@@ -0,0 +1,8 @@
...
+
+ENTRYPOINT ["cfn_nag"]
```
erickascic: maybe we need one for scan v. regular?

jesseadams: Not sure I follow

Dockerfile
jesseadams: I was going to do this when we figured out DockerHub and start publishing. This was just a starting point.

Dockerfile
erickascic: there are two command lines: cfn_nag and cfn_nag_scan.  the latter will search recursively through a directory structure for cfn templates and aggregate the results

jesseadams: We should be able to use both with the same Dockerfile, just default to one or the other, no?

PR #169, opened 2019-04-12
Refactoring for Metrics/MethodLength

jesseadams:
Acceptable method length <= 15 lines

### DIFF
--- a/lib/cfn-nag/cfn_nag.rb
+++ b/lib/cfn-nag/cfn_nag.rb
@@ -64,14 +64,14 @@ class CfnNag
     aggregate_results
   end
 
-  def violations_to_json(violations)
+  def audit_result(violations)
     {
       failure_count: Violation.count_failures(violations),
       violations: violations
     }
   end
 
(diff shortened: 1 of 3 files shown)


lib/cfn-nag/cfn_nag.rb
```diff
@@ -64,6 +64,19 @@ def audit_aggregate_across_files(input_path:,
...
   end
 
+  def violations_to_json(violations)
+    {
+      failure_count: Violation.count_failures(violations),
+      violations: violations
+    }
+  end
+
+  def generate_fatal_violation(message)
```
erickascic: just name this fatal_violation (drop the generate)

lib/cfn-nag/cfn_nag.rb
```diff
@@ -64,6 +64,19 @@ def audit_aggregate_across_files(input_path:,
     aggregate_results
   end
 
+  def violations_to_json(violations)
```
erickascic: maybe name this audit_result() instead?

PR #173, opened 2019-04-12
Refactoring for the following lint class cops

jesseadams:
* Lint/UnusedMethodArgument
* Lint/RescueException
* Lint/UnusedBlockArgument
* Lint/StringConversionInInterpolation


lib/cfn-nag/custom_rule_loader.rb
```diff
@@ -81,7 +81,7 @@ def filter_rule_classes(cfn_model, violations)
                                                                          allow_suppression: @allow_suppression
         audit_result = rule_class.new.audit(filtered_cfn_model)
         violations << audit_result unless audit_result.nil?
-      rescue Exception => exception
+      rescue StandardError => exception
```
erickascic: what will the difference in behavior be here?

lib/cfn-nag/jmes_path_discovery.rb
```diff
@@ -3,13 +3,13 @@ def initialize(rule_registry)
...
   end
 
-  def warning(id:, jmespath:, message:)
+  def warning(id:, message:)
     @rule_registry.definition(id: id,
                               type: Violation::WARNING,
                               message: message)
   end
 
-  def failure(id:, jmespath:, message:)
```
erickascic: i should just pull all the jmespath crud out.... it was an after thought and i don't really test it or think about it so it's pretty much dead code that distracts

jesseadams:
https://www.honeybadger.io/blog/ruby-exception-vs-standarderror-whats-the-difference/

PR #174, opened 2019-04-15
Fixing various rubocop violations

jesseadams:
Made exceptions for some

### DIFF
--- a/lib/cfn-nag/cfn_nag.rb
+++ b/lib/cfn-nag/cfn_nag.rb
@@ -82,8 +82,8 @@ class CfnNag
                                       parameter_values_string
       violations += @custom_rule_loader.execute_custom_rules(cfn_model)
       violations = filter_violations_by_profile violations
-    rescue Psych::SyntaxError, ParserError => parser_error
-      violations << fatal_violation(parser_error.to_s)
+    rescue Psych::SyntaxError, ParserError => exception
+      violations << fatal_violation(exception.to_s)
     rescue JSON::ParserError => json_parameters_error
       error = "JSON Parameter values parse error: #{json_parameters_error}"
       violations << fatal_violation(error)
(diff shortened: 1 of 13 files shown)


lib/cfn-nag/cfn_nag.rb
```diff
@@ -82,8 +82,8 @@ def audit(cloudformation_string:, parameter_values_string: nil)
                                       parameter_values_string
       violations += @custom_rule_loader.execute_custom_rules(cfn_model)
       violations = filter_violations_by_profile violations
-    rescue Psych::SyntaxError, ParserError => parser_error
-      violations << fatal_violation(parser_error.to_s)
+    rescue Psych::SyntaxError, ParserError => exception
```
erickascic: what is the value in changing this to have a less specific name?

lib/cfn-nag/custom_rules/CloudFormationAuthenticationRule.rb
```diff
@@ -19,15 +19,12 @@ def rule_id
...
     cfn_model.raw_model['Resources'].each do |resource_name, resource|
-      unless resource['Metadata'].nil?
-        unless resource['Metadata']['AWS::CloudFormation::Authentication'].nil?
-
-          resource['Metadata']['AWS::CloudFormation::Authentication'].each do |auth_name, auth|
-            if potentially_sensitive_credentials? auth
-              logical_resource_ids << resource_name
-            end
-          end
+      next if resource['Metadata'].nil?
```
erickascic: if we are going to spruce this up, let's use select() instead.

lib/cfn-nag/cfn_nag.rb
jesseadams: I think it is attempting to force consistency.

PR #179, opened 2019-04-23
Add CodeBuild::Project Missing EncryptionKey Warning

### LINKED ISSUE #69: Warn about unspecified EncryptionKey in AWS::CodeBuild::Project
not a fail because the default with encrypt with the AWS managed key but... just to bring this to the developer's attention

byronic:
Closes #69; Adds a Warning on no specified EncryptionKey for a CodeBuild::Project resource

### DIFF
--- a/lib/cfn-nag/custom_rules/CodeBuildEncryptionKeyRule.rb
+++ b/lib/cfn-nag/custom_rules/CodeBuildEncryptionKeyRule.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+
+# Rule class to ensure a CF distribution has logging
+class CodeBuildEncryptionKeyRule < BaseRule
+  def rule_text
+    'CodeBuild project should specify an EncryptionKey value'
+  end
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W32'
+  end
+
+  def audit_impl(cfn_model)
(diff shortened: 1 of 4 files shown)


lib/cfn-nag/custom_rules/CodeBuildEncryptionKeyRule.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+
+# Rule class to ensure a CF distribution has logging
```
erickascic: copy-pasta comment

lib/cfn-nag/custom_rules/CodeBuildEncryptionKeyRule.rb
```diff
@@ -0,0 +1,28 @@
...
+class CodeBuildEncryptionKeyRule < BaseRule
+  def rule_text
+    'CodeBuild project should specify an EncryptionKey value'
+  end
+
+  def rule_type
+    Violation::WARNING
```
erickascic: hmm... don't recall what the issue asks for but we usually treat missing encryption settings as a fail.  on the other hand, the default without a key is to use the amazon cmk which might be ok for some folks

lib/cfn-nag/custom_rules/CodeBuildEncryptionKeyRule.rb
byronic: d'oh

PR #182, opened 2019-04-23
Add SimpleAD Rule, DRY Refactor

### LINKED ISSUE #65: Check password in AWS::DirectoryService::SimpleAD for literal/noecho
like #49

byronic:
Closes #65; Adds a rule to enforce no hardcoded passwords, either in a parameter or directly in the DirectoryService::SimpleAD portion of the template; refactors some repeated private functions into a util directory

### DIFF
--- a/lib/cfn-nag/util/string_to_boolean.rb
+++ b/lib/cfn-nag/util/string_to_boolean.rb
@@ -1,7 +1,0 @@
-# frozen_string_literal: true
-
-# Migrated from multiple classes with no modifications.
-# I, @byronic, worry that this function isn't _quite_ what it says on the tin
-def to_boolean(string)
-  string.to_s.casecmp('true').zero?
-end
--- a/lib/cfn-nag/util/string_to_boolean.rb
+++ b/lib/cfn-nag/util/string_to_boolean.rb
@@ -1,0 +1,7 @@
+# frozen_string_literal: true
+
+# Migrated from multiple classes with no modifications.
+# I, @byronic, worry that this function isn't _quite_ what it says on the tin
+def to_boolean(string)
+  string.to_s.casecmp('true').zero?
+end
--- a/spec/util/enforce_noecho_parameter_spec.rb
+++ b/spec/util/enforce_noecho_parameter_spec.rb
@@ -1,0 +1,34 @@
+require 'spec_helper'
+require 'cfn-model'
+require 'cfn-nag/util/enforce_noecho_parameter'
+
+describe 'no_echo_parameter_without_default', :rule do
+
+  context 'test function no_echo_parameter_without_default' do
+    it 'returns false if there is a problem' do
+      cfn_model = CfnParser.new.parse read_test_template('json/rds_instance/rds_instance_no_echo_with_default_username.json')
+      cfn_model.resources_by_type('AWS::RDS::DBInstance')
+               .select do |instance|
+        if instance.masterUsername.nil?
+          fail "masterUsername shouldn't be nil"
+        else
+          expect(no_echo_parameter_without_default?(cfn_model, instance.masterUsername))
+            .to eq false
+        end
+      end
+    end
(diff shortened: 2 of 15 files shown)


lib/cfn-nag/util/string_to_boolean.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+# Migrated from multiple classes with no modifications.
+# I, @byronic, worry that this function isn't _quite_ what it says on the tin
+def to_boolean(string)
```
erickascic: yea this name is misleading per your worry.  this code is sort of a "weak" typing for "true", "True", "False", "false", true, false to be treated the same in comparison.

i'm not sure what the best name would be though... it could be as simple as is_true?().  you could indicate something in the name about how it handles string and boolean... that name might be pretty awkward: is_true_weak_typing?() lol.

another tact is to possibly name it after the context .... is_cfn_true? since cfn is the one treating all these string/boolean values the same when it comes to time to evaluate the resource

spec/util/string_to_boolean_spec.rb
```diff
@@ -0,0 +1,19 @@
+require 'spec_helper'
+require 'cfn-nag/util/string_to_boolean'
+
+describe 'to_boolean', :rule do
+
+  context 'Strings map to boolean' do
+    it 'given any true string, to_boolean returns true boolean' do
+      %w[true TRUE tRUE True tRuE TrUe]
```
erickascic: is it confirmed that cfn is cool with these other cases?  (i presume so, just checking)

spec/util/enforce_noecho_parameter_spec.rb
```diff
@@ -0,0 +1,34 @@
+require 'spec_helper'
+require 'cfn-model'
+require 'cfn-nag/util/enforce_noecho_parameter'
+
+describe 'no_echo_parameter_without_default', :rule do
+
+  context 'test function no_echo_parameter_without_default' do
+    it 'returns false if there is a problem' do
+      cfn_model = CfnParser.new.parse read_test_template('json/rds_instance/rds_instance_no_echo_with_default_username.json')
```
erickascic: totally cool to take the "sociable" unit testing approach here.... but Q: is there 100% coverage on the method (2 specs cover the 4 or so outcomes?)  if not, you can probably mock the model to get the last nook/cranny conditional...

spec/util/string_to_boolean_spec.rb
byronic: Update: It's not! Good catch. I was testing the function's tactic rather than the overarching strategy. Refactoring.

lib/cfn-nag/util/string_to_boolean.rb
byronic: Currently this also treats any non-truthy value (e.g. 'blue') as false. Is that the expected behavior?

If so, I'm cool with `is_true` or `is_truthy` as a renaming scheme.

spec/util/enforce_noecho_parameter_spec.rb
byronic: 80% coverage, but I can get it to 100. Will do

lib/cfn-nag/util/string_to_boolean.rb
jeffb4: (was clearing out my inbox)

`to_boolean` is not misleading - it attempts to convert a string value to a Boolean, just in a fairly naive manner ("blue" should raise a maybe an ArgumentError, and the method should probably handle "0"/"1", "yes", "no"). There's a good deal of prior art using this method name, and of course Ruby classes are full of `to_` methods.

An OO way to handle this that addresses Eric's note about context is to create a CfnString subclass of string and define a `.to_boolean` method that attempts to replicate CFN's rules for truthy string booleans.

PR #184, opened 2019-04-23
Hotfix RDS DBInstance With ClusterIdentifier

### LINKED ISSUE #183: Incorrect flagging of encrypted RDS clusters
When you set the AWS:RDS:DBCluster 'StorageEncrypted' property to 'true', it encrypts all instances in the cluster. cfn_nag is reporting a violation on each of the instances in the cluster, "RDS DBInstance should have StorageEncrypted enabled" even though this is not required when you configure an rds cluster.

byronic:
Closes #183; RDS database instances inherit the cluster's encryption status (if they are part of a cluster); fixes a false positive on database instance encryption in cases where a cluster identifier was present

### DIFF
--- a/lib/cfn-nag/custom_rules/RDSDBInstanceStorageEncryptedRule.rb
+++ b/lib/cfn-nag/custom_rules/RDSDBInstanceStorageEncryptedRule.rb
@@ -20,8 +20,11 @@ class RDSDBInstanceStorageEncryptedRule < BaseRule
     resources = cfn_model.resources_by_type('AWS::RDS::DBInstance')
 
     violating_instances = resources.select do |instance|
-      instance.storageEncrypted.nil? ||
-        instance.storageEncrypted.to_s.casecmp('false').zero?
+      instance.dBClusterIdentifier.nil? &&
+        (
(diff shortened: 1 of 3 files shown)


lib/cfn-nag/custom_rules/RDSDBInstanceStorageEncryptedRule.rb
```diff
@@ -20,8 +20,11 @@ def audit_impl(cfn_model)
     resources = cfn_model.resources_by_type('AWS::RDS::DBInstance')
 
     violating_instances = resources.select do |instance|
-      instance.storageEncrypted.nil? ||
-        instance.storageEncrypted.to_s.casecmp('false').zero?
+      instance.dBClusterIdentifier.nil? &&
+        (
+          instance.storageEncrypted.nil? ||
+          instance.storageEncrypted.to_s.casecmp('false').zero?
```
erickascic: we should have an empty? predicate for this to drop the CCM

erickascic:
BTW, a philosophical nit - we don't need no stinkin' "hot fixes".  all changes are equal and should be relatively small to cause a release as long as tests pass

Issue #149, opened 2019-01-07
Warn on resources with custom names


atkinsonm:
Cloned from https://gitlab.com/jhctechnology/aws-cloudformation-utilities/awslint/issues/19

Custom-named resources cannot be updated when the properties require replacement. This makes it more difficult to update those resources and adds complexity to a continuous deployment model.

erickascic:
latest cfn_nag warns on the following resource types + name properties:

'AWS::ApiGateway::ApiKey'                   => 'name',
    'AWS::CloudWatch::Alarm'                    => 'alarmName',
    'AWS::CodeDeploy::DeploymentConfig'         => 'deploymentConfigName',
    'AWS::CodeDeploy::DeploymentGroup'          => 'deploymentGroupName',
    'AWS::DynamoDB::Table'                      => 'tableName',
    'AWS::EC2::SecurityGroup'                   => 'groupName',
    'AWS::ECR::Repository'                      => 'repositoryName',
    'AWS::ElasticLoadBalancingV2::LoadBalancer' => 'name',
    'AWS::Elasticsearch::Domain'                => 'domainName',
    'AWS::IAM::Group'                           => 'groupName',
    'AWS::IAM::ManagedPolicy'                   => 'managedPolicyName',
    'AWS::IAM::Role'                            => 'roleName',
    'AWS::Kinesis::Stream'                      => 'name',
    'AWS::RDS::DBInstance'                      => 'dBInstanceIdentifier'

PR #189, opened 2019-05-01
Feature/warn_when_resource_has_explicit_name

seanmckinley:
This PR addresses issue #149, by warning when the resources specified in the issue (as well as a myriad of others) have a 'name' property explicitly set.

Admittedly a large portion of this PR is just tests and test data, hopefully I didn't go too overboard.


lib/cfn-nag/custom_rules/ResourceWithExplicitNameRule.rb
```diff
@@ -0,0 +1,59 @@
...
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W28'
+  end
+
+  def has_explicitly_set_resource_name?(resource, key_name)
```
erickascic: i'd make this guy private since only called internally

spec/custom_rules/profile.log
```diff
@@ -0,0 +1,2072 @@
+FUNCTION  <SNR>101_get_syn()
```
erickascic: is this committed by mistake?

seanmckinley: Oh lord, yeah it was. Good catch, let me remove it!

lib/cfn-nag/custom_rules/ResourceWithExplicitNameRule.rb
seanmckinley: Will do

seanmckinley:
@erickascic PR updated with suggestions, and a new test to ensure multiple offending resources within the same template are flattened the way we expect.

PR #197, opened 2019-05-09
Refactor CLI Interface, Add Scan fail-on-warnings

byronic:
[Finishes #165899506]; [Finishes #165899494]; factors out command line arguments reading; adds fail-on-warnings support to cfn_nag_scan; removes repeated code across cfn_nag_scan and cfn_nag by refactoring into an executor. Post-refactor, rubocop and testing required the following included supplmental changes: CfnNagExecutor is too large if read_cli_options is included, so factors that function out to its own file; CfnNag class is too large with the added fail_on_warnings option, refactors initialize parameters as a separate CfnNagConfig class; refactors the cfn_nag integration tests to use the new signatures


lib/cfn-nag/cfn_nag_executor.rb
```diff
@@ -0,0 +1,91 @@
...
+    unless opts[:blacklist_path].nil?
+      @blacklist_definition = IO.read(opts[:blacklist_path])
+    end
+
+    unless opts[:parameter_values_path].nil?
+      @parameter_values_string = IO.read(opts[:parameter_values_path])
+    end
+  end
+
+  def make_config(opts)
```
erickascic: i just call this method what it is without the verb: cfn_nag_config(opts)

lib/cfn-nag/read_cli_options.rb
```diff
@@ -0,0 +1,85 @@
+# frozen_string_literal: true
+
+require 'trollop'
+
+# rubocop:disable Metrics/BlockLength
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
+def read_cli_options(require_input_path: false)
```
erickascic: i'd just name this: cli_options()

lib/cfn-nag/cfn_nag_executor.rb
```diff
@@ -0,0 +1,93 @@
...
+class CfnNagExecutor
+  def initialize(require_input_path: false)
+    @require_input_path = require_input_path
+    @profile_definition = nil
+    @blacklist_definition = nil
+    @parameter_values_string = nil
+  end
+
+  def scan(aggregate_output: true,
+           cli_supplier: cli_options(cfn_nag_scan: @require_input_path),
```
erickascic: if i follow the code, cli_supplier and arg_supplier are exposed here to improve testability?  imho this is a tradeoff for complicating the interface - i.e. anybody who calls this might scratch their head and likely never change the defaults.  another approach i can suggest is to have "factory methods" for these two arguments, and then in your testing do a partial mock to replace the factory methods.  easy on the testing side, and nothing exposed to the consumer

lib/cfn-nag/cfn_nag_executor.rb
```diff
@@ -0,0 +1,93 @@
...
+
+class CfnNagExecutor
+  def initialize(require_input_path: false)
+    @require_input_path = require_input_path
+    @profile_definition = nil
+    @blacklist_definition = nil
+    @parameter_values_string = nil
+  end
+
+  def scan(aggregate_output: true,
```
erickascic: the require_input_path and aggregate_output are giving me a little trouble.  i know what they're there for.... but i'm wondering if you turn the other scan args into factory methods.... er do you need both? or could you have them both be on the scan() instead of split between the constructor and scan()?  there are basically two outcomes - single or aggregate, and only one of them requires the input path?

lib/cfn-nag/cli_options.rb
```diff
@@ -0,0 +1,143 @@
+# frozen_string_literal: true
+
+require 'trollop'
+
+# rubocop:disable Metrics/BlockLength
+# rubocop:disable Metrics/AbcSize
+# rubocop:disable Metrics/MethodLength
+def cli_options(cfn_nag_scan: false)
+  options_message = '[options] <cloudformation template path ...>|' \
```
erickascic: nit: move the messages into the "scope" where they are relevant and have the shared stuff up here only

lib/cfn-nag/cli_options.rb
```diff
@@ -0,0 +1,143 @@
...
+
+  template_pattern_message = 'Within the --input-path, match files to scan ' \
+                             'against this regular expression'
+
+  version = Gem::Specification.find_by_name('cfn-nag').version
+
+  if cfn_nag_scan
+    Trollop.options do
+      version version
+      opt :input_path,
```
erickascic: can we do one set of Trollop.options, but use the "cfn_nag_scan" flag to determine which short arg to force per opt for backward compat?

erickascic: also break the options into common vs. divergent.

lib/cfn-nag/argf.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+class Argf
```
erickascic: should be able to drop this whole file?

PR #199, opened 2019-05-10
Feature/end to end rspec tests

seanmckinley:
[Finishes #165899494]

    This commit introduces end-to-end tests for the cfn-nag gem, focusing on
    the cfn_nag binary itself. The local workflow leverages RVM to generate
    a new gemset, build and install cfn-nag locally using version 0.0.01,
    installing simplecov and rspec, and then executing tests tagged with
    'end_to_end'.

    The remote workflow (circleci) leverages a separate job and a fresh
    docker container to run the specs, and follows all of the same steps as
    the local workflow. The only real difference is that circleci does not
    leverage RVM, as these tests are already executing in isolation via the
    new docker container.


spec/end_to_end/cfn_nag_rules/rules_spec.rb
```diff
@@ -0,0 +1,18 @@
+require 'spec_helper'
+
+describe 'cfn_nag_rules', end_to_end: true do
+  context 'listing default installed rules' do
+    stdout,stderr,status = Open3.capture3("cfn_nag_rules")
```
erickascic: when do you use system() v Open3.capture3?

spec/end_to_end/cfn_nag/input_template_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'spec_helper'
+
+describe 'cfn_nag template file input', end_to_end: true do
+  context 'when reading in a template file with 1 failure' do
+    it "should print out 'failure_count': 1 " do
```
erickascic: nit: i'd scrub "should" from the it clauses.  like yoda would say, do - or do not, there is no try.  e.g. it "prints out failure count 1"

.circleci/config.yml
```diff
@@ -17,6 +17,14 @@ jobs:
       - run:
           name: Rspec tests
           command: bundle exec rspec --color --require spec_helper spec
+  end_to_end_test:
+    docker:
+      - image: circleci/ruby:2.5.5-browsers
+    steps:
+      - checkout
+      - run:
+          name: install cfn_nag-0.0.01 locally, rspec, pull sample templates, rspec test
```
erickascic: where is the 01 coming from? vs. 0.0.0?

seanmckinley:
> a few nits, a few questions.
> 
> i love the controlled approach to measuring the output of the CLI but if possible i'd also like to still include the original e2e tests that download the pile of cfn templates from aws. don't need assertions for them, just make sure they parse without exceptions.

Okay, that's a good call. I took that stuff out mostly because I lacked the context and wanted to speed the specs up. I'll add that back in though, since clearly that's a really good thing to test for.

spec/end_to_end/cfn_nag/version_spec.rb
```diff
@@ -0,0 +1,23 @@
+require 'spec_helper'
+
+describe 'cfn_nag --version', end_to_end: true do
+  context 'when ensuring the local gem is installed' do
+    it 'should equal 0.0.01' do
+      expect { system %( cfn_nag --version ) }
+        .to output(a_string_matching('0.0.01'))
```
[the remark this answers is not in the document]
erickascic: this seems reasonable given we have total control over the version on the 'in' side.  measuring the same string shows on the 'out' side.

spec/end_to_end/cfn_nag_rules/rules_spec.rb
erickascic: agreed system() reads a whole lot nicer.  makes sense.

PR #200, opened 2019-05-13
Fail on not utilizing NoEcho for MasterUserPassword in AWS::RDS::DBCluster

phelewski:
This creates a new Failing rule whenever the `MasterUserPassword` for RDS DBCluster is set in plaintext either in the resource itself, or as a default from a parameter. If `MasterUserPassword` for RDS DBCluster is specified, then it needs to be set as a parameter with `NoEcho`.

This takes care of [Story #164842091](https://www.pivotaltracker.com/story/show/164842091)

```
F32 RDS DB Cluster master user password must be Ref to NoEcho Parameter. Default credentials are not recommended
```

### DIFF
--- a/lib/cfn-nag/custom_rules/RDSDBClusterMasterUserPasswordRule.rb
+++ b/lib/cfn-nag/custom_rules/RDSDBClusterMasterUserPasswordRule.rb
@@ -1,0 +1,34 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require 'cfn-nag/util/enforce_noecho_parameter.rb'
+require_relative 'base'
+
+class RDSDBClusterMasterUserPasswordRule < BaseRule
+  def rule_text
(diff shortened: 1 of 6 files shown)


speshak:
I feel this incorrectly reports an error in a very specific application:       
```
MasterUserPassword: !If
  - RestoreSnapshot
  - !Ref 'AWS::NoValue'
  - !Ref Password
```

Where:
`Password` is a NoEcho stack parameter. 
`RestoreSnapshot` is a conditional that is true when `AWS::RDS::DBCluster` is given a `DBSnapshotIdentifier` parameter.  

When `DBSnapshotIdentifier` is present, `MasterUserPassword` must be empty.   The template complies with the spirit of this change, but it fails with `RDS DB Cluster master user password must be Ref to NoEcho Parameter. Default credentials are not recommended`

phelewski:
@speshak [0.4.17](https://rubygems.org/gems/cfn-nag/versions/0.4.17) from #228 now addresses and allows this use case scenario.

PR #206, opened 2019-05-25
S3BucketAccessLoggingRule

pethers:
S3 buckets should define access logging.


erickascic:
Hey thanks for the PR.  Please see #209 for details on what I had to add.  

1. The class name was a copy of another rule class
2. The test referenced a test template that wasn't there (extra 'in')
3. Added a positive test
4. A good number of unit tests failed after introducing this rule - the regression tests weren't expecting W35 so they would fail.  Updated older test templates to include LoggingConfiguration to avoid breaking the unit tests.

PR #216, opened 2019-05-30
Start of moving rules into separate gem.

jesseadams:
Depends on https://github.com/stelligent/cfn-nag-rules-core getting published

* Load rules from ObjectSpace instead of file paths
* Created new base class: CfnNag::BaseRule
* Moved WorkspacesWorkspaceEncryptionRule to CfnNag::Workspaces::WorkspaceEncryptionRule


lib/cfn-nag/custom_rule_loader.rb
```diff
@@ -193,12 +195,17 @@ def discover_rule_classes(rule_directory)
 
     rule_filenames.each do |rule_filename|
       require(rule_filename)
-      rule_classname = File.basename(rule_filename, '.rb')
+    end
 
-      rule_classes << Object.const_get(rule_classname)
+    ObjectSpace.each_object do |object|
```
erickascic: i'm not usually one to care *so* much about performance but.... this iterates over every object in the heap?  what kind of performance price do we pay for this? (or can we pass BaseRule to filter?)

spec/custom_rules/base_spec.rb
```diff
@@ -44,14 +42,12 @@ def rule_text
...
 
-        dontcare = double('cfn_model')
-
         expected_violation = Violation.new(id: 'F3333',
                                            type: Violation::FAILING_VIOLATION,
                                            message: 'This is an epic fail!',
                                            logical_resource_ids: %w[r1 r2 r3])
 
-        expect(base_rule.audit(dontcare)).to eq expected_violation
+        expect(base_rule.audit(nil)).to eq expected_violation
```
erickascic: reason for getting rid of "dontcare"? (which shows more intention)

lib/cfn-nag/custom_rule_loader.rb
erickascic: is this looking for Class objects somehow v. any/every object instance?

jesseadams: This is looking at all objects in the heap. I'll look into ways to increase performance. I think rule detection may be happening multiple times so one optimization may be to cache the rules found the first time and keep using them.

spec/custom_rules/base_spec.rb
jesseadams: After my changes rspec was complaining about this double leaking into other specs. Since we are not really using the double features at all, nil seems to be an equivalent and the tests pass.

lib/cfn-nag/custom_rule_loader.rb
jesseadams: @erickascic performance has been addressed.

spec/custom_rules/base_spec.rb
erickascic: that's cool - i'd set a variable named dontcare = nil just to make it clear that the argument has nothing to do with the outcome of the test

jesseadams:
> Does this affect custom rule loading in any way? i.e. the extra rule directory argument still works ok?

This is currently backwards compatible.

> Do we have to move over all the current "core" rules before merging this?

No. We can slowly move rules over and add new rules to the gem as we see fit.

> What is the UX for this.... users install the latest cfn-nag, they get the latest rules. if the upgrade nag, they get the latest rules, or... they can just upgrade the rules gem?

I think we'd pin the rules gem in cfn-nag with a `>=` operator so folks will get at least the bare minimum version we dictate, but can separately upgrade the underlying rules.

PR #221, opened 2019-06-03
Prevent duplicate ids

### LINKED ISSUE #220: Prevent duplicate rule IDs from occurring
Right now it is possible to commit two rules with the same IDs and all tests will pass and the gem will be published.

jesseadams:
Closes #220

### DIFF
--- a/lib/cfn-nag/rule_registry.rb
+++ b/lib/cfn-nag/rule_registry.rb
@@ -3,10 +3,15 @@
 require_relative 'rule_definition'
 
 class RuleRegistry
-  attr_reader :rules
+  attr_reader :rules, :duplicate_ids
 
   def initialize
     @rules = []
+    @duplicate_ids = []
+  end
+
+  def duplicate_ids?
+    @duplicate_ids.count.positive?
   end
 
   def definition(id:,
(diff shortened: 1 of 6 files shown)


lib/cfn-nag/rule_registry.rb
```diff
@@ -20,7 +25,11 @@ def definition(id:,
     if existing_def.nil?
       add_rule rule_definition
     else
-      existing_def
+      @duplicate_ids << {
+        id: id,
+        new_message: message,
```
erickascic: the new v. old is fine, but poking at two things:
1. is the order stable for this to have meaning?
2. on the off chance we have > 2 collisions... i guess it shows A/B and then B/C?

Perhaps make @duplicate_ids a Hash with id key and array of messages?

I guess there's no way to say *where* the duplicate class is coming from?  for core rules this is a bug..... but when there are custom rule directories that users are pointing to....

jesseadams: I have suggested that custom rules have their own number system with a prefix of `C`. But yeah, this is just a barebones approach to fail CI builds when there are duplicates. The messages aren't perfect but they enabled me to quickly troubleshoot the one duplicate we had. If we want something fancier, we can always iterate on it in the future as need arises.

PR #227, opened 2019-06-12
Remove need for RVM

kidbrax:
This runs the tests through Docker, eliminating the need for RVM

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,16 @@
+bundle:
+	docker run \
+		--rm \
+		--volume $$(pwd):/usr/src/app \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -0,0 +1,16 @@
...
+		ruby:2.6 \
+		bundle install
+
+test:
+	docker run \
+		--tty \
+		--rm \
+		--volume $$(pwd):/usr/src/app \
+		--workdir /usr/src/app \
+		ruby:2.6 \
```
jesseadams: We currently support ruby 2.5.x

Issue #201, opened 2019-05-14
False Positive: Rule failing when pulling from secrets manager


paul-nelson-baker:
It looks like there was a pull about an hour or so ago. It immediately blocks our RDS cluster that uses SecretsManager directly and doesn't use parameters, but uses [dynamic references](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/dynamic-references.html).

[Here is the change that did it](https://github.com/stelligent/cfn_nag/commit/b48878ba61221737f2922d9be756a190311b2f7c)

And here is a pseudo snippet of what our template looks like

```yaml
  RDSDatabaseCluster:
    Type: AWS::RDS::DBCluster
    Properties:
      DBSubnetGroupName: !Ref RDSSubnetGroup 
      DBClusterIdentifier: !Ref RDSClusterIdentifier
      Engine: aurora
      EngineMode: provisioned
      StorageEncrypted: true
      KmsKeyId: !ImportValue GenericRDSKey
      MasterUsername: '{{resolve:secretsmanager:/DBA/rds/masteruser:SecretString:username}}'
      MasterUserPassword: '{{resolve:secretsmanager:/DBA/rds/masteruser:SecretString:password}}'
      VpcSecurityGroupIds:
        - !Ref RDSSecurityGroup
      DBClusterParameterGroupName: !Ref RDSClusterParameterFamilyAurora
```

erickascic:
hey sorry we'll take a look at this tout de suite.  in the mean, you can suppress the offending rule with the --blacklist-path (or the --profile-path white list feature)

paul-nelson-baker:
Mucho appreciado! Thanks for your looking at this 😄

erickascic:
0.3.92 has reverted this rule.  we've got some work to do on dealing with the secrets references.... as a caveat, the rds dbinstance (and maybe neptune and redshift? not sure) might also have this issue.

phelewski:
[0.4.17](https://rubygems.org/gems/cfn-nag/versions/0.4.17) adds this ability to allow a Dynamic Reference for Secrets Manager for the `MasterUserPassword`.

PR #228, opened 2019-06-12
Fail on not utilizing NoEcho for MasterUserPassword in AWS::RDS::DBCluster

phelewski:
This adds back in the code from the old PR #200 

> This creates a new Failing rule whenever the MasterUserPassword for RDS DBCluster is set in plaintext either in the resource itself, or as a default from a parameter. If MasterUserPassword for RDS DBCluster is specified, then it needs to be set as a parameter with NoEcho.

But this time it adds in a test case and adjusted rule for allowing a dynamic reference to Secrets Manager.

This takes care of Issue #201 

```
F34 RDS DB Cluster master user password must be Ref to NoEcho Parameter. Default credentials are not recommended
```


lib/cfn-nag/util/enforce_secrets_manager.rb
```diff
@@ -0,0 +1,19 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/util/truthy.rb'
```
jesseadams: This require is unnecessary and should be removed.

lib/cfn-nag/util/enforce_secrets_manager.rb
```diff
@@ -0,0 +1,19 @@
...
+    if key_to_check.start_with? '{{resolve:secretsmanager:'
+      if key_to_check.end_with? '}}'
+        return true
+      end
+    else
+      return false
+    end
+  end
+  false
+end
```
jesseadams: I am surprised rubocop is OK with these nested ifs. I would assume it would prefer something like:

```
def secrets_manager_property_value?(_cfn_model, key_to_check)
  return false unless key_to_check.is_a? String
  return false unless key_to_check.start_with? '{{resolve:secretsmanager:'
  return false unless key_to_check.end_with? '}}'

  true
end
```

I don't feel super strongly either way

jesseadams:
Tests pass, but rubocop needs re-enabled for the class in question and this needs refactored/cleaned up a bit.

jesseadams:
This is ready for review. May be good to merge as is but we may want to eventually rip out some of this into a library if we are going to parse conditionals in other rules in the future.

lib/cfn-nag/util/enforce_secrets_manager.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+# Returns true if the provided key_to_check is a dynamic reference
+# to Secrets Manager; false otherwise.
+def secrets_manager_property_value?(_cfn_model, key_to_check)
+  return false unless key_to_check.is_a? String
+  return false unless key_to_check.start_with? '{{resolve:secretsmanager:'
```
erickascic: can we also deal with other dynamic references like ssm and ssm-secure?

lib/cfn-nag/custom_rules/RDSDBClusterMasterUserPasswordRule.rb
```diff
@@ -0,0 +1,83 @@
...
+    parameter_value =
+      cfn_model.parameters[ref].synthesized_value ||=
+        cfn_model.parameters[ref].default
+    password_true = cluster.masterUserPassword['Fn::If'][1]
+    password_false = cluster.masterUserPassword['Fn::If'][2]
+
+    [password_true, password_false, parameter_value]
+  end
+
+  def conditional_is_invalid?(cfn_model, cluster)
```
erickascic: i'm not sure this accounts for nested conditionals and other nasties that can be embedded within the output values for a conditional.....

phelewski:
@erickascic The conditional logic is removed now. Care to review again?

lib/cfn-nag/custom_rules/RDSDBClusterMasterUserPasswordRule.rb
```diff
@@ -24,6 +24,8 @@ def audit_impl(cfn_model)
     violating_rdsclusters = rds_dbclusters.select do |cluster|
       if cluster.masterUserPassword.nil?
         false
+      elsif cluster.masterUserPassword.is_a?(Hash) &&
+            cluster.masterUserPassword.key?('Fn::If')
```
jesseadams: Think we need a `false` here, or combine this conditional with the one above

lib/cfn-nag/custom_rules/RDSDBClusterMasterUserPasswordRule.rb
```diff
@@ -0,0 +1,40 @@
...
+    rds_dbclusters = cfn_model.resources_by_type('AWS::RDS::DBCluster')
+    violating_rdsclusters = rds_dbclusters.select do |cluster|
+      if cluster.masterUserPassword.nil?
+        false
+      else
+        !no_echo_parameter_without_default?(cfn_model,
+                                            cluster.masterUserPassword) &&
+          !dynamic_reference_property_value?(cfn_model,
+                                             cluster.masterUserPassword) &&
+          !conditional_intrinsic_func_present?(cfn_model,
```
erickascic: definitely don't want to flag intrinsic functions but i'd hunt down what was causing the conditions to be flagged as a violation.  i believe we can just look for the known bad stuff and ignore everything else (like conditions).

lib/cfn-nag/util/enforce_reference_parameter.rb
```diff
@@ -0,0 +1,20 @@
...
+  return false unless key_to_check.key? 'Ref'
+
+  if cfn_model.parameters.key? key_to_check['Ref']
+    parameter = cfn_model.parameters[key_to_check['Ref']]
+    if truthy?(parameter.noEcho) && parameter.default.nil?
+      return false
+    end
+  end
+
+  true
```
erickascic: this is dead code?

lib/cfn-nag/util/enforce_dynamic_reference.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
+
+# Returns false if the provided key_to_check is a dynamic reference
+# to SSM Secure or Secrets Manager; true otherwise.
+# Only applicable for a string
+def insecure_dynamic_reference?(_cfn_model, key_to_check)
```
erickascic: i think this logic is correct - but the name of the method seems to state that it is only checking dynamic references.... perhaps a better name is insecure_literal_or_dynamic_reference?()

lib/cfn-nag/util/enforce_reference_parameter.rb
erickascic: nvm - i see it isn't dead code... the default being !nil will fall through here.

erickascic: so small nit would either be to comment the fallthrough, or else make the conditions explicit like if X then Y else Z to make it clear what the branches are

phelewski: I've added more comments into this function to hopefully make things more clear from a quick glance.

lib/cfn-nag/util/enforce_dynamic_reference.rb
phelewski: naming has been changed

lib/cfn-nag/custom_rules/DirectoryServiceSimpleADPasswordRule.rb
```diff
@@ -1,7 +1,8 @@
 # frozen_string_literal: true
 
 require 'cfn-nag/violation'
-require 'cfn-nag/util/enforce_noecho_parameter'
+require 'cfn-nag/util/enforce_string_or_dynamic_reference'
+require 'cfn-nag/util/enforce_string_or_dynamic_reference'
```
jesseadams: This is required twice?

PR #229, opened 2019-06-18
converted Make to Rake and added a few new commands

kidbrax:
previously Make only allowed for running e2e tests


Rakefile
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+docker_run_prefix = "docker run --tty --rm --volume #{Dir.pwd}:/usr/src/app --workdir /usr/src/app cfn-nag-dev:latest"
+
+def dev_image_missing?
+  image = `docker images -q cfn-nag-dev:latest`
+
+  if image.empty?
+    puts 'The local dev image is missing. You must first run `rake build_docker_dev`'
```
jesseadams: Why not just automatically invoke this task?

Rakefile
```diff
@@ -0,0 +1,40 @@
...
+desc 'Build the local Docker iamge for development/testing'
+task :build_docker_dev do
+  sh 'docker build --file Dockerfile-dev --rm --tag cfn-nag-dev .'
+end
+
+namespace 'test' do
+  desc 'Run all rspec tests'
+  task :all do
+    ensure_local_dev_image
+    sh "#{docker_run_prefix} ./scripts/rspec.sh"
```
erickascic: it makes sense to me to run e2e tests in a docker container in an installed gem... it seems counter-intuitive to me to run my unit tests against an installed gem vs just running them in my dev environment against the actual source base... i mean i guess it's not wrong?  just seems like extra over what i need to get fast feedback during development

kidbrax: you could still get fast feedback. everything just runs in a local container vs directly on your machine. The idea was to not have to install anything locally (in case i'm working on several projects where i may have several versions of ruby, python, node, etc required). The commands/tests should be exactly the same though

Issue #66, opened 2017-10-18
Check password in AWS::DirectoryService::MicrosoftAD for literal/noecho


erickascic:
like #49

erickascic:
https://www.pivotaltracker.com/story/show/164842081

twellspring:
Closed by #235

PR #235, opened 2019-06-21
Fail on not utilizing NoEcho for Password in AWS::DirectoryService::MicrosoftAD

phelewski:
This takes care of Issue #66 

This creates a new Failing rule whenever the Password for Directory Service Microsoft Active Directory is set in plaintext either in the resource itself, or as a default from a parameter. If Password for Directory Service Microsoft Active Directory is specified, then it needs to be set as a parameter with NoEcho.

```
F36 DirectoryService::MicrosoftAD password must be Ref to NoEcho Parameter. Default credentials are not recommended
```

### DIFF
--- a/lib/cfn-nag/custom_rules/DirectoryServiceMicrosoftADPasswordRule.rb
+++ b/lib/cfn-nag/custom_rules/DirectoryServiceMicrosoftADPasswordRule.rb
@@ -1,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require 'cfn-nag/util/enforce_reference_parameter'
+require 'cfn-nag/util/enforce_string_or_dynamic_reference'
+require_relative 'base'
+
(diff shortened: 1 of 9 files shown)


lib/cfn-nag/custom_rules/DirectoryServiceMicrosoftADPasswordRule.rb
```diff
@@ -0,0 +1,35 @@
...
+
+require 'cfn-nag/violation'
+require 'cfn-nag/util/enforce_reference_parameter'
+require 'cfn-nag/util/enforce_string_or_dynamic_reference'
+require_relative 'base'
+
+# Rule class to fail on DirectoryService::MicrosoftAD password in template
+class DirectoryServiceMicrosoftADPasswordRule < BaseRule
+  def rule_text
+    'DirectoryService::MicrosoftAD password must be Ref to NoEcho Parameter. ' \
```
erickascic: same comment as for DMS

Issue #64, opened 2017-10-18
Check Password in AWS::DMS::Endpoint for literal/Noecho


erickascic:
Like #49

erickascic:
https://www.pivotaltracker.com/story/show/164842083

twellspring:
Closed by #236

PR #236, opened 2019-06-21
Fail on not utilizing NoEcho for Password in AWS::DMS::Endpoint

phelewski:
This takes care of Issue #64 

This creates a new Failing rule whenever the Password for DMS Endpoint is set in plaintext either in the resource itself, or as a default from a parameter. If Password for DMS Endpoint is specified, then it needs to be set as a parameter with NoEcho.

```
F37 DMS Endpoint password must be Ref to NoEcho Parameter. Default credentials are not recommended
```

### DIFF
--- a/lib/cfn-nag/custom_rules/DMSEndpointPasswordRule.rb
+++ b/lib/cfn-nag/custom_rules/DMSEndpointPasswordRule.rb
@@ -7,8 +7,8 @@ require_relative 'base'
 
 class DMSEndpointPasswordRule < BaseRule
   def rule_text
-    'DMS Endpoint password must be Ref to NoEcho Parameter. ' \
-    'Default credentials are not recommended'
+    'DMS Endpoint must not be a plaintext string or a Ref to a NoEcho ' \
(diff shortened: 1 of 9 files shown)


lib/cfn-nag/custom_rules/DMSEndpointPasswordRule.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require 'cfn-nag/util/enforce_reference_parameter'
+require 'cfn-nag/util/enforce_string_or_dynamic_reference'
+require_relative 'base'
+
+class DMSEndpointPasswordRule < BaseRule
+  def rule_text
+    'DMS Endpoint password must be Ref to NoEcho Parameter. ' \
```
erickascic: let's flip this rule_text around.  DMS endpoint must not be plaintext string of a Ref to a Noecho parm with default.  we're letting the dynamic references by or anything else we don't "understand"

Issue #238, opened 2019-06-24
Consolidate password rule logic into a base rule


jesseadams:
Right now we have several password rules that duplicate logic. Create a base rule with the logic and have all implementing rule classes use it.

PR #239, opened 2019-06-24
Create consolidated Password rule

phelewski:
Takes care of issue #238 

Simplifies existing and future password rules by creating a new base password rule. This cuts away a lot of the duplicate coding needed for each rule. Now you only need to supply the `resource_type` & `password_property` and the new base password rule will automatically check if the resource property value is `nil` or insecure.

I've gone ahead and updated each of the current password rules that were already utilizing the `insecure_parameter` & `insecure_string_or_dynamic_reference` methods.

Also, now there is a `password_rule_spec_helper.rb` to help with creating spec tests for the password rules. This eliminates a lot of the duplicate code needed to run specs against the 8 test templates for each rule.


lib/cfn-nag/custom_rules/password_base_rule.rb
```diff
@@ -0,0 +1,32 @@
...
+  def password_property
+    raise 'must implement in subclass'
+  end
+
+  def audit_impl(cfn_model)
+    resources = cfn_model.resources_by_type(resource_type)
+
+    violating_resources = resources.select do |resource|
+      if resource.send(password_property).nil?
+        false
```
erickascic: it's probably a harmless "pass through" but the password property isn't optional for all of these resources.  it is optional for resources that can come from a snapshot, but the various directory resources require it

jesseadams: May be a good idea to create a `password_required` that defaults to true that can be overridden in the implementing subclass

PR #243, opened 2019-06-26
#147 Feature/147

### LINKED ISSUE #147: Fail on iam:PassRole with resource *
Cloned from https://gitlab.com/jhctechnology/aws-cloudformation-utilities/awslint/issues/21

`iam:PassRole` on `Resource: "*"` can be dangerous because it can allow a user to elevate their own permissions by passing a privileged role to a trusted entity to run commands they could not normally run on their own.

This is an extension of existing cfn_nag rules W11, W12, and W13 to escalate these warnings to failures when a policy contains the `iam:PassRole` action.

twellspring:
Fixes #147 

All tests passing.  Please let me know if the file naming makes sense ... there seemed to be multiple naming conventions.


twellspring:
Need to get linter/formatter working.  I'll update this once that is done.

twellspring:
Got Rubocop installed/working for formatting/linting.  Updates tested.

lib/cfn-nag/custom_rules/IamManagedPolicyPassRoleWildcardResourceRule.rb
```diff
@@ -0,0 +1,36 @@
...
+
+  def rule_type
+    Violation::FAILING_VIOLATION
+  end
+
+  def rule_id
+    'F40'
+  end
+
+  def passrole_action?(statement)
```
erickascic: private

lib/cfn-nag/custom_rules/IamManagedPolicyPassRoleWildcardResourceRule.rb
```diff
@@ -0,0 +1,36 @@
...
+    statement.actions.find { |action| action == 'iam:PassRole' }
+  end
+
+  def wildcard_resource?(statement)
```
erickascic: private

lib/cfn-nag/custom_rules/IamManagedPolicyPassRoleWildcardResourceRule.rb
```diff
@@ -0,0 +1,36 @@
...
+  def rule_type
+    Violation::FAILING_VIOLATION
+  end
+
+  def rule_id
+    'F40'
+  end
+
+  def passrole_action?(statement)
+    statement.actions.find { |action| action == 'iam:PassRole' }
```
erickascic: i'm not sure we've ever written a rule before that spells out particular actions?  i'm thinking we need to "relax" this equality check s.t. it matches up against a partial wildcard.  for example, iam:Pass*, iam:PassR* should match.  there's probably something generic we can write, but for now, a naive regex match is good enough?

lib/cfn-nag/custom_rules/IamPolicyPassRoleWildcardResourceRule.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+
+class IamPolicyPassRoleWildcardResourceRule < BaseRule
```
erickascic: consider deriving this from a base class to share the code with managedpolicy

lib/cfn-nag/custom_rules/IamManagedPolicyPassRoleWildcardResourceRule.rb
twellspring: Fixed

lib/cfn-nag/custom_rules/IamManagedPolicyPassRoleWildcardResourceRule.rb
twellspring: done

lib/cfn-nag/custom_rules/IamManagedPolicyPassRoleWildcardResourceRule.rb
twellspring: Created a iam_permutations utility.

lib/cfn-nag/custom_rules/IamPolicyPassRoleWildcardResourceRule.rb
twellspring: Base class created

lib/cfn-nag/util/iam_permutations.rb
```diff
@@ -0,0 +1,50 @@
+# frozen_string_literal: true
+
+# Find all pemutations of an IAM string with wildcards approprate to the location
+
+def iam_permutations(string, element, prefix: '', suffix: '')
```
erickascic: i'd say everything looks good in this PR except for this guy.  most important review comment is that this is very error prone code that lends itself wonderfully to automated unit tests..... but i don't see any, so that's my main gripe.

i have some advice on the interface and the implementation.  

for the interface.... it's not clear to me why the "action", prefix or suffix are needed.  i'd expect to pass in a string like PassRole and get the collection of wildcards back (and then up to me to add iam: if I so choose.  On the other hand, wildcards behave differently by *service*... but i wouldn't sweat that just yet.

I honestly didn't try to decipher the implementation in great depth since it didn't have the tests to read first.  that said, i don't know for sure, but instead of saying "i bet" - let's say - "it wouldn't surprise me if...." a recursive solution would be about half the code and read more concisely.

twellspring: I converted it to recursive and made it a generic `wildcard_patterns`.  adding the `iam:` at the front is taken care of now with a map.

To give wildcard_patterns flexibility I have an optional `patterns` parameter with the default being to include wildcards before, after and both before and after.

twellspring: Recursive is definitely much clearer / less code.

lib/cfn-nag/custom_rules/passrole_base_rule.rb
```diff
@@ -0,0 +1,35 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+require 'cfn-nag/util/wildcard_patterns'
+
+class PassRoleBaseRule < BaseRule
+  IAM_ACTION_PATTERNS = wildcard_patterns('PassRole').map! { |x| 'iam:' + x } + ['*']
```
erickascic: two small nits here:

1. i don't believe you need to use map! vs. just map? or is there some technicality with this being a constant?  bang sometimes means there is a whacky side effect 
2. rename "x" as pass_role_with_asterisk or the like for the readability

lib/cfn-nag/util/wildcard_patterns.rb
```diff
@@ -0,0 +1,43 @@
...
+  return results if string.empty?
+
+  results << "*#{string}"
+  wildcard_front(string[1..-1], results)
+end
+
+def wildcard_front_back(string, results = [])
+  return results if string.empty?
+
+  results += wildcard_back(string.to_s, prepend: '*')
```
erickascic: is the to_s necessary given string is a String?  if it is necessary, provide a comment for the why

twellspring: the to_s was needed because passing in an integer causes `.empty?` to throw an exception.  So  I moved the type conversion to a higher level and changed code to better document ( why comment if code is clear).

lib/cfn-nag/custom_rules/passrole_base_rule.rb
twellspring: Put the ! because I thought that was needed to change the original array, but yeah thinking about it the assigning the results means the ! is not needed.   

Changed name to `pattern` which is consistent with naming in other locations using a map on a plural variable.

PR #246, opened 2019-06-27
Feature/148

### LINKED ISSUE #148: Warn on missing security group or rule descriptions
Cloned from https://gitlab.com/jhctechnology/aws-cloudformation-utilities/awslint/issues/17 and https://gitlab.com/jhctechnology/aws-cloudformation-utilities/awslint/issues/18

Security groups and security group rules without a description obscure their purpose and may lead to bad practices in ensuring they only allow traffic from the ports and sources/destinations required.

twellspring:
Fixes #148

### DIFF
--- a/lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
+++ b/lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
@@ -1,0 +1,54 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+require 'cfn-nag/ip_addr'
+
+class SecurityGroupRuleDescriptionRule < BaseRule
+  def rule_text
+    'Security group rules without a description obscure their purpose and may'\
+    'lead to bad practices in ensuring they only allow traffic from the ports'\
+    'and sources/destinations required.'
+  end
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W36'
+  end
+
+  def violating_security_groups?(cfn_model)
+    violating_security_groups = cfn_model.security_groups.select do |security_group|
+      violating_ingres = security_group.ingresses.select do |ingress|
+        ingress.description.nil? && ingress.logical_resource_id.nil?
+      end
+      violating_egress = security_group.egresses.select do |egress|
+        egress.description.nil? && egress.logical_resource_id.nil?
+      end
+      !violating_ingres.empty? || !violating_egress.empty?
+    end
+    violating_security_groups.map(&:logical_resource_id)
+  end
+
+  def violating_ingress?(cfn_model)
+    violating_ingress = cfn_model.resources_by_type('AWS::EC2::SecurityGroupIngress').select do |standalone_ingress|
+      standalone_ingress&.description.nil?
+    end
+    violating_ingress.map(&:logical_resource_id)
+  end
+
+  def violating_egress?(cfn_model)
+    violating_egress = cfn_model.resources_by_type('AWS::EC2::SecurityGroupEgress').select do |standalone_ingress|
+      standalone_ingress&.description.nil?
+    end
+    violating_egress.map(&:logical_resource_id)
+  end
+
+  def audit_impl(cfn_model)
+    violating_security_groups?(cfn_model) +
+      violating_ingress?(cfn_model) +
+      violating_egress?(cfn_model)
+  end
+end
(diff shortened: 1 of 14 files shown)


lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W36'
+  end
+
+  def violating_sg_part(sg_part)
```
erickascic: make all of these violating* methods private

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
```diff
@@ -0,0 +1,54 @@
...
+
+  def violating_ingress?(cfn_model)
+    violating_ingress = cfn_model.resources_by_type('AWS::EC2::SecurityGroupIngress').select do |standalone_ingress|
+      standalone_ingress&.description.nil?
+    end
+    violating_ingress.map(&:logical_resource_id)
+  end
+
+  def violating_egress?(cfn_model)
+    violating_egress = cfn_model.resources_by_type('AWS::EC2::SecurityGroupEgress').select do |standalone_ingress|
```
erickascic: copy-pasta - should be |egress|

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
```diff
@@ -0,0 +1,54 @@
...
+  def violating_security_groups?(cfn_model)
+    violating_security_groups = cfn_model.security_groups.select do |security_group|
+      !violating_sg_part(security_group.ingresses).empty? || !violating_sg_part(security_group.egresses).empty?
+    end
+    violating_security_groups.map(&:logical_resource_id)
+  end
+
+  def violating_ingress?(cfn_model)
+    violating_ingress = cfn_model.resources_by_type('AWS::EC2::SecurityGroupIngress').select do |standalone_ingress|
+      standalone_ingress&.description.nil?
```
erickascic: this safe navigation operator shouldn't be necessary.  if ingress.description is raising an exception, then there is something we have to fix in cfn-model.  the ModelElement.method_missing() should return nil for any property that isn't found.

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
```diff
@@ -0,0 +1,54 @@
...
+    end
+    violating_ingress.map(&:logical_resource_id)
+  end
+
+  def violating_egress?(cfn_model)
+    violating_egress = cfn_model.resources_by_type('AWS::EC2::SecurityGroupEgress').select do |standalone_ingress|
+      standalone_ingress&.description.nil?
```
erickascic: it's not enough to check nil.  have to check for an empty string after trimming

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
```diff
@@ -0,0 +1,54 @@
...
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W36'
+  end
+
+  def violating_sg_part(sg_part)
+    sg_part.select do |item|
+      item.description.nil? && item.logical_resource_id.nil?
```
erickascic: what is this checking?  an inline rule won't have a logic resource id.

i was confused at first by the word "part".   Maybe call the method inline_rules_have_missing_description?(rules)

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
twellspring: fixed

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
twellspring: fixed

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
twellspring: hmmm ... it was consistently giving me an error.  But just took it out and it is working now. So I must have been doing something different when I got the error.   

Resolved

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
twellspring: fixed

lib/cfn-nag/custom_rules/SecurityGroupRuleDescriptionRule.rb
twellspring: Fixed

PR #255, opened 2019-07-05
Create EBSEncryptionRule.rb

annasaphira:
Created new rule to check if EBS is encrypted.

### DIFF
--- a/lib/cfn-nag/custom_rules/EBSEncryptionRule.rb
+++ b/lib/cfn-nag/custom_rules/EBSEncryptionRule.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'boolean_base_rule'
+
+class EBSEncryptionRule < BooleanBaseRule
+  def rule_text
+    'EBS Volume should use true for property encryption'
+  end
+  
+  def rule_type
+    Violation::FAILING_VIOLATION
+  end
+
+  def rule_id
+    'F38'
+  end
+
+  def resource_type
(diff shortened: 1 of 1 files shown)


jesseadams:
Is this any different than Rule ID `F1` - https://github.com/stelligent/cfn_nag/blob/master/lib/cfn-nag/custom_rules/EbsVolumeHasSseRule.rb ?

I realize that `EbsVolumeHasSseRule` does need updated to use `BooleanBaseRule`, but otherwise I think it is the same.

annasaphira:
Ok - that's right - we overlooked that - sorry - so you can use it as an update for the other rule. My team and I discuss another rule, that should check if parameter encryption_key has a value. Do you know if you have this in your development pipe?

erickascic:
@annasaphira i don't believe we have an existing issue in our backlog for anything related to EC2::Volume. KmsKeyId.

There are a couple of options for how we could help you....

I'm not sure if we want to emit a failing violation for having no KmsKeyId.  Some ppl want the basic SSE and don't want to pay for and to manage KMS keys.... that said, we could either:

1. Add a rule to the core product where it is a warning, or...
2. We could write the rule as a violation and give you the code that you can use with a custom profile.

annasaphira:
Thank you for your reply, we would prefer your first proposition. 
Can you tell us, when it will be released?

jesseadams:
@annasaphira - Closing this in favor of https://github.com/stelligent/cfn_nag/issues/261

We will try to get this release within a week. Thanks for your patience.

erickascic:
Released in 0.4.33

PR #257, opened 2019-07-05
#242 - Add new 'colortxt' output_format, 'txt' output_format no longer colorizes output

### LINKED ISSUE #242: Disable colorize?
So the colorized output is cool for interactive runs, but makes reviewing Jenkins logs a little problematic.  Could we get a flag to disable colorization, i.e. when running cfn_nag as part of a pipeline?

jesseadams:
Closes #242

### DIFF
--- a/lib/cfn-nag/cfn_nag_executor.rb
+++ b/lib/cfn-nag/cfn_nag_executor.rb
@@ -69,7 +69,7 @@ class CfnNagExecutor
   end
 
   def validate_options(opts)
-    unless opts[:output_format].nil? || %w[txt json].include?(opts[:output_format])
+    unless opts[:output_format].nil? || %w[colortxt txt json].include?(opts[:output_format])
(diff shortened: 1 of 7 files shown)


lib/cfn-nag/cfn_nag_executor.rb
```diff
@@ -69,7 +69,7 @@ def scan_file(cfn_nag, fail_on_warnings)
   end
 
   def validate_options(opts)
-    unless opts[:output_format].nil? || %w[txt json].include?(opts[:output_format])
+    unless opts[:output_format].nil? || %w[colortxt txt json].include?(opts[:output_format])
       Trollop.die(:output_format,
                   'Must be txt or json')
```
kidbrax: shouldn't this include colortxt now?

erickascic: yea the die message should include colortxt

PR #266, opened 2019-07-22
This will convert output of cfn_nag to csv

munntjlx:
This would take the output of cfn_nag_rules and output to csv. Its a bit pedantic but it works.

### DIFF
--- a/scripts/cfnrules2csv.pl
+++ b/scripts/cfnrules2csv.pl
@@ -1,0 +1,7 @@
+#!/usr/local/bin/perl -w
+while (<>) {
+  $correctedvariable = $_;
+  $correctedvariable =~ s/(^\S+)/$1\,/g;
+  $correctedvariable =~ s/, /,/g;
+  print $correctedvariable;
+}


jesseadams:
I think this may best be implemented as an official `output-format` of `csv`.

munntjlx:
I would be happy to do that. Will submit later today or tomorrow.  I am assuming from stdin, so I can use a foreach loop instead of a while if you prefer.  Not sure how to directly integrate with ruby, as I am a ruby novice. The regexes are the important parts, I expect.

Sincerely,

Thomas J. Munn
Consulting Security Engineer
LexisNexis


On 7/22/19, 14:41, "Mr-Lizard" <notifications@github.com<mailto:notifications@github.com>> wrote:


*** External email: use caution ***



@Mr-Lizard commented on this pull request.

Works nicely - short, simple, sweet ! One suggestion though, since it's a utility to specifically convert cfn_nag_rules output, you could make the program do that directly by using

foreach (cfn_nag_rules) {

For your loop instead of the while.

twellspring:
@munntjlx, thanks for your work on this. I agree with @jesseadams that we should put this in code. I was looking at adding more output formats due to another user's question. 

I have created https://github.com/stelligent/cfn_nag/issues/269
I'll be working on this next week

twellspring:
@munntjlx Closing this PR because #269 was just merged which provides csv and json formatted output.

PR #267, opened 2019-07-25
#86 install cfn_nag using brew on Mac and Linux

### LINKED ISSUE #86: Add ability to install through homebrew

twellspring:
Closes #86 

Update Readme to include brew gem installation method.

@erickascic The brew tap solution would require homebrew-tap/Formula/cfn-nag.rb to be updated any time a dependency of cfn-nag.gemspec gets updated. We could make a script that reads cfn-nag.gemspec and then writes a new homebrew-tap/Formula/cfn-nag.rb but that means having to handle the `~>` and seems messy and prone to errors.  The `brew-gem` does not ensure ruby is installed, but that is easy enough to add to the install instructions:

```
brew install ruby brew-gem
brew gem install cfn-nag
```

This appears to be the simplest from a maintenance standpoint and it is two commands just like the tap version would be, so the complexity for users is basically the same.

### DIFF
--- a/README.md
+++ b/README.md
@@ -25,11 +25,11 @@ gem install cfn-nag
 ```
 
 ## Brew Install
-On MacOS or Linux (tested on ubuntu) install with brew
+On MacOS or Linux you can alternatively install with brew
 
 ```bash
-brew tap stelligent/homebrew-tap
-brew install cfn-nag
+brew install ruby brew-gem
+brew gem install cfn-nag
 ```
 
 # Pipeline
--- a/README.md
+++ b/README.md
@@ -26,7 +26,8 @@ gem install cfn-nag
 
 ## Brew Install
 On MacOS or Linux (tested on ubuntu) install with brew
-```
+
+```bash
 brew tap stelligent/homebrew-tap
 brew install cfn-nag
 ```
--- a/README.md
+++ b/README.md
@@ -17,10 +17,18 @@ For more background on the tool, please see:
 
 # Installation
 
-Presuming Ruby 2.5.x is installed, installation is just a matter of `
+## Gem Install 
+Presuming Ruby >= 2.5.x is installed, installation is just a matter of `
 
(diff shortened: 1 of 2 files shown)


twellspring:
Hold on this PR. I totally removed cfn_nag (and all dependencies) and brew from my system and re-installed brew, then tried to install cfn_nag via brew and it did not work.

Mr-Lizard:
Are you by any chance using rbenv ?  I'm noticing that it doesn't seem to play well with brew.

twellspring:
@Mr-Lizard yes I use rbenv

I just created a clean MacOS on Virtualbox and the install worked fine.  

The issue we have now is setting up automation to update the brew definition every time cfn_nag is updated. An alternative I am looking at is https://github.com/sportngin/brew-gem

With that the install of cnf_nag would be
```# brew install brew-gem
# brew gem install cfn_nag```

And then no automation is needed to keep the homebrew-tap definition up to date

Mr-Lizard:
Well that would *totally undo* all my work on the formula in question.  But...who cares ?  It would work beautifully and be trouble free!

twellspring:
@Mr-Lizard Thanks for your understanding on this. The more I dug in to automation of the updating of the homebrew-tap definition and having to hard code in the exact versions of every dependency, the less I liked the solution.  I played with parsing the cfn_nag gemspec file and building the homebrew-tap based on the dependencies listed in the gemspec.  But that seems messy and error prone.

twellspring:
Once this is merged, I will PR the removal of the tap definition.

Mr-Lizard:
@twellspring , No problem.  It is very decent of you to give me credit for this PR, thank you.   Have you evaluated the 'brew gem' method anymore ?

twellspring:
@Mr-Lizard  Yes, I updated this PR to use the 'brew gem' method.  Just waiting for it to be approved.

Issue #77, opened 2017-10-18
Replace godawful publish.sh with a readable Ruby version


erickascic:
if you pick this up, I have some code to start with

erickascic:
https://www.pivotaltracker.com/story/show/164842070

juddmon:
PR https://github.com/stelligent/cfn_nag/pull/365

PR #274, opened 2019-07-31
#77 Convert publish.sh script to ruby

twellspring:
Replace godawful publish.sh with a readable Ruby version

**This PR needs testing**
The publish.rb has specs and passes those tests, but there is currently no environment for testing it in an end-to-end manner.


twellspring:
Close ... accidentally put another issue on this PR. I will open 2 new PRs

Issue #273, opened 2019-07-30
New rule to warn on ipProtocol -1


atkinsonm:
Setting `IpProtocol: '-1'` can produce unexpected results. Per the [docs](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-security-group-rule.html#cfn-ec2-security-group-rule-ipprotocol), when `-1` is used, `ToPort` and `FromPort` are essentially ignored and access is granted on all ports from all protocols.

I believe this configuration should be treated the same as W27 and throw a warning for port ranges instead of single ports.

Credit to @grolston for the original find.

twellspring:
@atkinsonm Thanks for the input. Yes having a warning for this makes sense.

Mr-Lizard:
I have to ask - what is meant by a 'dangling' egress/ingress ?

Mr-Lizard:
created PR for this - [# 279](https://github.com/stelligent/cfn_nag/pull/279)

erickascic:
> I have to ask - what is meant by a 'dangling' egress/ingress ?

the egress/ingress is "dangling" if there is a resource for an ingress or egress in a given cfn template, but the security group it attaches to is NOT in the same template.  for example template1 could define a security group.  the template2 could accept an sg-id as a Parameter and create an egress/ingress rule using that sg-id.

Mr-Lizard:
Thanks, I couldn't find a definition anywhere for that.

atkinsonm:
Thanks all!

PR #279, opened 2019-08-08
#273 - New rule to warn on ipProtocol -1

Mr-Lizard:
Ok,let's get this party started !


lib/cfn-nag/custom_rules/SecurityGroupEgressPortRangeRule2.rb
```diff
@@ -0,0 +1,38 @@
...
+    'W40'
+  end
+
+  ##
+  # This will behave slightly different than the legacy jq based rule which was
+  # targeted against inline ingress only
+  def audit_impl(cfn_model)
+    violating_security_groups = cfn_model.security_groups.select do |security_group|
+      violating_egresses = security_group.egresses.select do |egress|
+        egress.ipProtocol == -1 && egress.fromPort != egress.toPort
```
erickascic: so i'm trying to grok this version of the rule vs. the existing rule.  the existing rule warns on any port ranges.... this warns when a rule has all protocols and a port range, so it's going to double up the violations.

i guess the question to ask is whether it should always be a warning where all protocols are allowed?  my gut at this point is to make this rule about just ipProtocol, and leave the port range to the other rule.

on a technical note - i think you need to make sure -1 and "-1" are take care of?

given it's just about the protocol, i''d change the name of the class to reflect that -  SecurityGroupEgressAllProtocolsRule

Mr-Lizard: Well, I wasn't happy about the name myself, so I'm glad a more succinct one has been suggested.

Will change code to just regard ipProtocol.

So all port ranges in general are considered suspicious ?

erickascic: yea... "suspicious" might be too strong a word but.... it's a warning to make sure you meant to open up that many ports.  there are legit reasons to do it, so it's not a failing violation....

Mr-Lizard: Ah, good.  I knew some clients have to open up the ephemeral port range, because the server would reply on a random port.  But I had no idea how servers behaved.

Mr-Lizard:
@erickascic, changes ready for review

Issue #75, opened 2017-10-18
Look for AdministratorAccess managed policy in IAM Role


erickascic:
If IAM::Role has ManagedPolicyArns that include AdministratorAccess, or any other uber powerful AWS managed policies...this should at least be a warning

melusom:
I'd also add `PowerUserAccess` and `IAMFullAccess` to that list. Rule suppression will allow users to ignore this warning for certain resources, but I think we can generally agree that this is not the recommended config and should be called out.

erickascic:
https://www.pivotaltracker.com/story/show/164842072

PR #293, opened 2019-10-15
Iam AdministratorAccess managed policy rule

mcahill7:
PR for Look for AdministratorAccess managed policy in IAM Role #75
https://github.com/stelligent/cfn_nag/issues/75


lib/cfn-nag/custom_rules/IamRoleAdministratorAccessPolicyRule.rb
```diff
@@ -0,0 +1,29 @@
...
+  def rule_text
+    'IAM role should not have AdministratorAccess policy'
+  end
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W25'
```
erickascic: W25 used to be something i forgot... probably something with security groups that got collapsed.  can you bump this to something higher? W38 maybe?

lib/cfn-nag/custom_rules/IamRoleAdministratorAccessPolicyRule.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def audit_impl(cfn_model)
+    violating_roles = cfn_model.resources_by_type('AWS::IAM::Role').select do |role|
+      violating_policies = role.managedPolicyArns.select do |policy|
```
erickascic: this is fine, but you could just do a "find" instead - quit after the first one since it only takes one to violate

spec/test_templates/json/iam_role/iam_role_with_administratoraccess_policy.json
```diff
@@ -0,0 +1,23 @@
...
+            "Statement": [ {
+              "Effect": "Allow",
+              "Principal": {
+                "Service": [ "ec2.amazonaws.com" ],
+                "AWS" : "arn:aws:iam::324320755747:root"
+              },
+              "Action": ["sts:AssumeRole"]
+            } ]
+          },
+          "ManagedPolicyArns" : [ "arn:aws:iam::aws:policy/AdministratorAccess" ]
```
erickascic: this is likely good enough given how much other testing there is against IAM roles, but generally i would suggest also including a "positive" test whereby your prove the rule/violation _won't_ be triggered.

PR #294, opened 2019-10-18
Modify Rule W43 to include warning for PowerUserAccess and IAMFullAccess

mcahill7:
Modify Rule W43 to include warning for PowerUserAccess and IAMFullAccess

https://github.com/stelligent/cfn_nag/issues/75


Mr-Lizard:
Is there an issue this corresponds to ?

lib/cfn-nag/custom_rules/IamRoleElevatedManagedPolicyRule.rb
```diff
@@ -0,0 +1,29 @@
...
+  end
+
+  def rule_id
+    'W43'
+  end
+
+  def audit_impl(cfn_model)
+    violating_roles = cfn_model.resources_by_type('AWS::IAM::Role').select do |role|
+      role.managedPolicyArns.find do |policy|
+        policy.include?('arn:aws:iam::aws:policy/AdministratorAccess') ||
```
erickascic: if you are still available to work on this.... i'd prefer not to delete/replace a rule that has already been out in the wild.  there is a "provenance" issue from changing what rules do.  like an artifact pushed to a binary repo - rules should be immutable beyond bug fixes. i would keep W43 as is, then add this "elevated privilege" rule to include power user and iam full access.  rest of it looks good though.

one small nit - i'd create a "predicate" function that takes a policy and spits out yea/nay so that the rule invokes it vs. a long disjunctive expression

Issue #253, opened 2019-07-02
Create remaining Password property rules


phelewski:
Add in the missing Resource/Property rules for passwords, tokens, and keys.

Current missing items:
- AWS::Amplify::App.BasicAuthConfig
  - [x] Password
- AWS::DMS::Endpoint.MongoDbSettings
  - [x] Password
- AWS::AppStream::DirectoryConfig.ServiceAccountCredentials
  - [x] AccountPassword
- AWS::OpsWorks::Stack.RdsDbInstance
  - [x] DbPassword
- AWS::Amplify::Branch.BasicAuthConfig
  - [x] Password
- AWS::IAM::User.LoginProfile
  - [x] Password
- AWS::OpsWorks::App.SslConfiguration
  - [x] PrivateKey
- AWS::OpsWorks::Stack.Source
  - [x] Password
- AWS::EMR::Cluster.KerberosAttributes
  - [x] ADDomainJoinPassword
  - [x] CrossRealmTrustPrincipalPassword
  - [x] KdcAdminPassword
- AWS::AmazonMQ::Broker.User
  - [x] Password
- AWS::KinesisFirehose::DeliveryStream.RedshiftDestinationConfiguration
  - [x] Password
- AWS::OpsWorks::App.Source
  - [x] Password
- AWS::KinesisFirehose::DeliveryStream.SplunkDestinationConfiguration
  - [x] HECToken
- ~~Alexa::ASK::Skill.AuthenticationConfiguration~~
  - ~~[ ] ClientSecret~~ Save this for another issue as it doesn't follow the `AWS::` format
- ~~Alexa::ASK::Skill.AuthenticationConfiguration~~
  - ~~[ ] RefreshToken~~ Save this for another issue as it doesn't follow the `AWS::` format
- AWS::CodePipeline::Webhook.WebhookAuthConfiguration
  - [x] SecretToken
- AWS::Amplify::App
  - [x] OauthToken
  - [x] AccessToken
- ~~AWS::CloudFormation::CustomResource~~
  - ~~[ ] ServiceToken~~ Not a secured value
- AWS::Pinpoint::APNSChannel
  - [x] TokenKey
  - [x] PrivateKey
- AWS::Pinpoint::APNSSandboxChannel
  - [x] TokenKey
  - [x] PrivateKey
- AWS::ElastiCache::ReplicationGroup
  - [x] AuthToken
- AWS::Pinpoint::APNSVoipSandboxChannel
  - [x] TokenKey
[...]

PR #298, opened 2019-10-24
Adding custom rules and tests for issue #253 - AWS::IAM::User.LoginPr…

tmcelhattan:
Adding custom rules and tests for issue #253 - AWS::IAM::User.LoginProfile password property rules


lib/cfn-nag/custom_rules/IamUserLoginProfilePasswordResetRule.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require "cfn-nag/violation"
+require_relative "base"
+
+class IamUserLoginProfilePasswordResetRule < BaseRule
+  def rule_text
+    "IAM User Login Profile should exist and have PasswordResetRequired property set to true"
```
erickascic: make sure to use single quotes if you don't need interpolation.

lib/cfn-nag/custom_rules/IamUserLoginProfilePasswordResetRule.rb
```diff
@@ -0,0 +1,26 @@
...
+    Violation::WARNING
+  end
+
+  def rule_id
+    "W50"
+  end
+
+  def audit_impl(cfn_model)
+    violating_iam_users = cfn_model.resources_by_type("AWS::IAM::User").select do |iam_user|
+      iam_user.loginProfile["PasswordResetRequired"].nil? || iam_user.loginProfile["PasswordResetRequired"].to_s == "false"
```
erickascic: make sure to use the proper truthy? predicate here.  cfn is "weakly typed" when it comes to boolean values

a few things:
1. this code is going to fail when loginProfile isn't there.  it's not a required field
2. i'm not sure what the proper behavior is here.... i guess if there is a login profile, the password should be noecho and reset must be required?

tmcelhattan:
@erickascic  Also made some additional refactors here as well based on the MQbroker PR.

Issue #299, opened 2019-10-28
Rule for api gateway access logging enabled


pethers:
Validate that AccessLogSetting(https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigateway-deployment-stagedescription.html) contains a value for DestinationArn and Format.

Probably have time to implement it myself, but just wanted to create an issue as well.

PR #300, opened 2019-10-28
Rule for api gateway access logging enabled https://github.com/stelligent/cfn_nag/issues/299

pethers:


erickascic:
released in 0.4.24

PR #301, opened 2019-10-30
Feature/issue 253 amazonmq broker user password

tmcelhattan:
Adding custom rules and tests for issue #253 - AWS::AmazonMQ::Broker User Password property rules.


lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
```diff
@@ -0,0 +1,40 @@
...
+
+  def rule_type
+    Violation::FAILING_VIOLATION
+  end
+
+  def rule_id
+    'F52'
+  end
+
+  def resource_type
```
erickascic: this method is unnecessary since this class doesn't derive from PasswordBaseRule

lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
```diff
@@ -0,0 +1,40 @@
...
+    'AWS::AmazonMQ::Broker'
+  end
+
+  def audit_impl(cfn_model)
+    resources = cfn_model.resources_by_type('AWS::AmazonMQ::Broker')
+
+    violating_resources = resources.select do |resource|
+      violating_users = resource.users.select do |user|
+        insecure_parameter?(cfn_model, user['Password']) || \
```
erickascic: 1. break this logic out into a predicate.  the backslashes give me the willies, but aside from that a complex boolean like this deserves its own named method to lower CCM and improve readability

2. checking the nil.... i'm hemming and hawing over that.  cfn-model is the underlying gem that is supposed to parse and prepare things so that we don't have to do a lot of this checking in nag rules.  since this is a nested data structure though (User) we are stuck doing some of that work here.  i guess i'd suggest checking it's there before trying to reference it... but then if it's not there, just do nothing.

spec/custom_rules/AmazonMQBrokerUserPasswordRule_spec.rb
```diff
@@ -0,0 +1,101 @@
...
+      )
+      actual_logical_resource_ids =
+        AmazonMQBrokerUserPasswordRule.new.audit_impl cfn_model
+      expected_logical_resource_ids = %w[]
+
+      expect(actual_logical_resource_ids).to eq expected_logical_resource_ids
+    end
+  end
+
+  context 'AmazonMQBroker has password from Not Secure String in Systems Manager' do
```
erickascic: missing password is one of the contexts to capture for coverage

lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
```diff
@@ -0,0 +1,40 @@
...
+        insecure_string_or_dynamic_reference?(cfn_model, user['Password']) || \
+        user['Password'].nil?
```
erickascic: actually.... can't you derive form PasswordBaseRule and use the subproperty behavior?  i'm not for sure on that

lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def check_user_property(user_property)
+    # Checks to make sure that the 'Users' property that is part of the 'AWS::AmazonMQ::Broker'
+    # resource exists and is defined.
+    if !user_property.nil?
+      true
+    end
+  end
+
+  def get_violating_users(cfn_model, resource)
```
erickascic: call this violating_users?  name the resource argument as"mq_broker

lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
```diff
@@ -0,0 +1,67 @@
...
+        true
+      elsif user['Password'].nil?
+        true
+      end
+    else
+      true    
+    end
+  end
+
+  def check_user_property(user_property)
```
erickascic: i'd nuke this method.  the return values could be true or nil which is possibly confusing - but you can check the nil? inline where this is invoked imho

lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
```diff
@@ -0,0 +1,67 @@
...
+
+  def rule_type
+    Violation::FAILING_VIOLATION
+  end
+
+  def rule_id
+    'F52'
+  end
+
+  def check_password_property(cfn_model, user)
```
erickascic: 'check' is a bit opaque as a term.  while i would never say comments are never necessary, imho writing them usually means something isn't obvious from the code itself.

user_has_insecure_password? is probably a better name

lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
erickascic: make this private

lib/cfn-nag/custom_rules/AmazonMQBrokerUserPasswordRule.rb
erickascic: make this private

tmcelhattan:
@erickascic Made some additional refactors based on your comments:
1. Removed that `check_user_property` method
2. Renamed the methods to make what they perform more clear
3. Made those renamed class methods private.

Issue #308, opened 2019-11-15
missing access logging rule for AWS::ApiGatewayV2


pethers:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-apigatewayv2-stage-accesslogsettings.html

pethers:
Started but wip at https://github.com/pethers/cfn_nag/commit/3abd1241e81ba64ac5add3094cda7fde1b54e001

PR #312, opened 2019-11-22
ApiGateway V2 should have access logging configured

pethers:
https://github.com/stelligent/cfn_nag/issues/308


pethers:
New rule : ApiGateway V2 should have access logging configured

Issue #311, opened 2019-11-17
Missing access logging rule for AWS::ElasticLoadBalancingV2


pethers:

pethers:
Started but wip at https://github.com/pethers/cfn_nag/commit/3abd1241e81ba64ac5add3094cda7fde1b54e001

erickascic:
included in 0.4.51.

PR #313, opened 2019-11-22
Missing access logging rule for AWS::ElasticLoadBalancingV2

pethers:
https://github.com/stelligent/cfn_nag/issues/311


pethers:
New rule : Elastic Load Balancer V2 should have access logging enabled

pethers:
Updated the string compare to ignore case, don't think weak typing is used in the case of string compare.

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-elasticloadbalancingv2-loadbalancer-loadbalancerattributes.html

access_logs.s3.enabled - Indicates whether access logs are enabled. The value is true or false. The default is false.

PR #319, opened 2019-12-04
Feature/create remaining password rules

phelewski:
Closes Issue #253 

This change request handles the remaining AWS Resource Properties that contain sensitive information, mainly passwords and secrets.

The list of rules that were created can be found in this comment box, https://github.com/stelligent/cfn_nag/issues/253#issue-463244821

The rules that were created utilized the password helpers to minimize duplicate code.
* A new password helper rule was created to work with sub property items that were in List form, `lib/cfn-nag/custom_rules/password_base_rule_sub_property_with_list.rb`. 
* The existing helper file, `lib/cfn-nag/custom_rules/password_base_rule.rb` was slightly tweaked to ensure that cfn_nag wouldn't fail when a property was defined in the rule, but not present in the CloudFormation Template itself.

A few already existing password rules were updated to use the new password helpers, these were:
* AWS::AmazonMQ::Broker.User Password
* AWS::IAM::User.LoginProfile Password
* AWS::OpsWorks::Stack.RdsDbInstance DbPassword

One of the remaining password properties didn't fit with the existing password helper rule without needing too much work to get it adjusted, if needed in the future this can be added with a new issue.
* AWS::ManagedBlockchain::Member.MemberFabricConfiguration AdminPassword


phelewski:
@erickascic I'll go and split up the PR into smaller per-service chunks. 

As for Amazon MQ Broker, I updated it to use `Users` since that is what the actual property name is, https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-amazonmq-broker.html#cfn-amazonmq-broker-users

Issue #314, opened 2019-11-22
Topic should specify a KmsMasterKeyId value


pethers:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-sns-topic.html

erickascic:
in 0.4.53

PR #321, opened 2019-12-04
Feature/issue 314 sns topic kmsmasterkeyid - Issue #314

tmcelhattan:
Checks to see if the `KmsMasterKeyId` property is defined for a SNS Topic resource and flags a warning violation if it is not. For Issue #314


erickascic:
released in 0.4.53

PR #323, opened 2019-12-09
Feature/add secure rule for alexa ask skill authentication configuration

### LINKED ISSUE #322: Create secure rules for Alexa::ASK::Skill.AuthenticationConfiguration
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ask-skill-authenticationconfiguration.html

Alexa::ASK::Skill.AuthenticationConfiguration
-  [x] ClientSecret
-  [x] RefreshToken

phelewski:
Closes #322 

Creating two new rules for `Alexa::ASK::Skill.AuthenticationConfiguration`
- ClientSecret
- RefreshToken

These rules are using the `BaseRule` class. New specs are created to test familiar template scenarios.

### DIFF
--- a/lib/cfn-nag/custom_rules/AlexaASKSkillAuthenticationConfigurationRefreshTokenRule.rb
+++ b/lib/cfn-nag/custom_rules/AlexaASKSkillAuthenticationConfigurationRefreshTokenRule.rb
@@ -1,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require 'cfn-nag/util/enforce_reference_parameter'
+require 'cfn-nag/util/enforce_string_or_dynamic_reference'
+require_relative 'base'
+
+class AlexaASKSkillAuthenticationConfigurationRefreshTokenRule < BaseRule
+  def rule_text
+    'Alexa ASK Skill AuthenticationConfiguration RefreshToken must not be ' \
+    'a plaintext string or a Ref to a NoEcho Parameter with a Default value.'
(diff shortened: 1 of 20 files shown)


lib/cfn-nag/custom_rules/AlexaASKSkillAuthenticationConfigurationRefreshTokenRule.rb
```diff
@@ -0,0 +1,36 @@
...
+  end
+
+  def rule_id
+    'F75'
+  end
+
+  def audit_impl(cfn_model)
+    ask_skills = cfn_model.resources_by_type('Alexa::ASK::Skill')
+    violating_skills = ask_skills.select do |skill|
+      refresh_token = skill.authenticationConfiguration['RefreshToken']
```
erickascic: one thing to be wary of - authenticationConfiguration is required but you could face a malformed template.  haven't been totally consistent about handling this.  cfn-model can hide/catch some of these issues up front so rule developers don't have to think about it..... you could add a kwalify schema into cfn-model, or you could check missing authenticationConfiguration here.

erickascic: if somebody jams an Fn::If at the top-level of properties, cfn-model is going to ignore the FnIf, but then the resource will have no properties.  there are a lot of places that are going to possibly break with that.... so not something you nec need to care about.

PR #324, opened 2019-12-09
Creating and Updating password helpers and updating opsworks stack rds password rule

phelewski:
Partial for #253 

* Updating the `password_base_rule` helper rule to check if the property exists within the CloudFormation template first.
* Creating a new `password_base_rule_sub_property_with_list` helper rule to use with resources that have a sub property in a `List` form.
* Updating existing `AWS::OpsWorks::Stack.RdsDbInstances DbPassword` rule to use newly created `PasswordBaseRuleSubPropertyWithList` helper class.
* Updating existing `OpsWorksStackRdsDbInstancePasswordRule` class to use the full AWS resource and property name.


lib/cfn-nag/custom_rules/password_base_rule.rb
```diff
@@ -36,7 +36,9 @@ def audit_impl(cfn_model)
 private
 
 def verify_parameter_exists(resource, password_property, sub_property_name)
-  if sub_property_name.nil?
+  if resource.send(password_property).nil?
+    true
```
erickascic: i read this as:
Given the password_property does not exist on the resource
Then the "parameter" exists

So i'm wondering if there are two naming aspects we need to fix:
1. "parameter" is really "password_property"?
2. The other names might be confusing me, but is it really a matter of existence?  If the password_property returns a nil value, I guess I'd expect "exists?" predicate to be false?

lib/cfn-nag/custom_rules/password_base_rule_sub_property_with_list.rb
```diff
@@ -0,0 +1,60 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require 'cfn-nag/util/enforce_reference_parameter'
+require 'cfn-nag/util/enforce_string_or_dynamic_reference'
+require_relative 'base'
+
+class PasswordBaseRuleSubPropertyWithList < BaseRule
```
erickascic: small nit.  when i subclass something, I typically preserve that parent class' name as the suffix, e.g. XBaseRule < BaseRule.

phelewski: I can fix this to be `PasswordSubPropertyWithListBaseRule`

lib/cfn-nag/custom_rules/password_base_rule.rb
phelewski: 1. I think I can tweak the naming here for the parameter/sub_property_name to work a bit better and read more clearly. The original setup was to just have the first level parameter from `Properties`, but there are several "parameters" now that are in a sub level even further down. 

2. I see the confusing naming and results now that you point that out. I need to address this to read more clearly so it makes more sense from a quick glance.

lib/cfn-nag/custom_rules/password_base_rule.rb
```diff
@@ -31,32 +31,34 @@ def audit_impl(cfn_model)
...
-def verify_parameter_exists(resource, password_property, sub_property_name)
-  if sub_property_name.nil?
-    resource.send(password_property).nil?
-  else
-    resource.send(password_property)[sub_property_name].nil?
+  def property_does_not_exist(resource, password_property, sub_property_name)
+    if resource.send(password_property).nil?
+      true
+    elsif sub_property_name.nil?
+      resource.send(password_property).nil?
```
erickascic: if we have fallen through to here, don't we know that password_property == nil?  in combination isn't the result true if password_property or sub_property == nil?

phelewski: If we make it to line 40 here then it means that the `password_property` does exist, but if the `sub_property_name` == nil? then we can return `false` since the property **would** exist.

I'll push a change so that we don't have to check `resource.send(password_property).nil?` twice here.

lib/cfn-nag/custom_rules/password_base_rule.rb
```diff
@@ -31,32 +31,34 @@ def audit_impl(cfn_model)
...
-def verify_parameter_exists(resource, password_property, sub_property_name)
-  if sub_property_name.nil?
-    resource.send(password_property).nil?
-  else
-    resource.send(password_property)[sub_property_name].nil?
+  def property_does_not_exist(resource, password_property, sub_property_name)
+    if resource.send(password_property).nil?
+      true
+    elsif sub_property_name.nil?
+      false
```
erickascic: lol - yes i think this is correct! but the not_exist really turns things upside down

erickascic:
released in 0.4.55.  had to fix a few small rubocop nits too

Issue #60, opened 2017-10-18
Add rules for AWS::ElasticLoadBalancingV2::Listener ssl policy


erickascic:
We could be opinionated about this and force TLS 1.1 and 1.2, or.... we could define check per policy and let the user add the particular rules to a profile

erickascic:
https://www.pivotaltracker.com/story/show/164842087

thegonch:
PR created at: https://github.com/stelligent/cfn_nag/pull/354

PR #354, opened 2020-01-31
Feature/60 elbv2 listener ssl policy rules

thegonch:
Address issue 60:  https://github.com/stelligent/cfn_nag/issues/60

- Adds rule to warn when using an SSL Policy for an ELB v2 Listener that is not for TLS 1.2
- Adds rule to warn when using a Protocol of HTTP for an ELB v2 Listener (since we should be generally using HTTPS for an ALB where possible)
- Adds applicable spec tests and test template files

### DIFF
--- a/lib/cfn-nag/custom_rules/ElasticLoadBalancerV2ListenerProtocolRule.rb
+++ b/lib/cfn-nag/custom_rules/ElasticLoadBalancerV2ListenerProtocolRule.rb
@@ -3,9 +3,9 @@
 require 'cfn-nag/violation'
 require_relative 'base'
 
-class ElasticLoadBalancerV2ListenerSslPolicyRule < BaseRule
+class ElasticLoadBalancerV2ListenerProtocolRule < BaseRule
   def rule_text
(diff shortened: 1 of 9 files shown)


pethers:
Nice rule, looks like it detects all pre tls 1.2 policies 

ELBSecurityPolicy-2016-08 (default)
ELBSecurityPolicy-TLS-1-0-2015-04
ELBSecurityPolicy-TLS-1-1-2017-01
ELBSecurityPolicy-FS-2018-06
ELBSecurityPolicy-FS-1-1-2019-08
ELBSecurityPolicy-2015-05

from https://docs.aws.amazon.com/elasticloadbalancing/latest/application/create-https-listener.html

gustav-sundin-wcar:
Hi! Just wanted to point out that the correct casing according to the CloudFormation docs is `SslPolicy`, not `SSLPolicy` as in the .yml files in this PR. Also, in ElasticLoadBalancerV2ListenerSslPolicyRule.rb you check for `sSLPolicy`. Not sure if this makes a difference or if it works anyway.

Issue #67, opened 2017-10-18
Investigate whether MfaConfiguration=ON for AWS::Cognito::UserPool is a sane check


erickascic:
Will require understanding if there are any use cases where we really want it OFF or OPTIONAL..... might be a warning if we usually want ON, but can have a reason for other

erickascic:
https://www.pivotaltracker.com/story/show/164842080

PR #366, opened 2020-02-07
Initial commit for Issue #67 - Adding rule to check for Cognito UserP…

tmcelhattan:
Initial commit for Issue #67  
- Adding rule to check for Cognito UserPool MfaConfiguration set to 'ON' or 'OPTIONAL' . 
- Added tests for different variations of accepted keyword values in CloudFormation template for `AWS::Cognito::UserPool` resource.

Notes:
1. It seems reasonable that users who build Cognito UserPools can set the sign up process to either enforce MFA configuration (`MfaConfiguration: 'ON'`) or have it optional for people who either have MFA or do not (`MfaConfiguration: 'OPTIONAL'`). It seems like a good idea to WARN on MFA configuration being off though, (`MfaConfiguration: 'OFF'`).
2. The reason for this is that there are cases where a user might use Cognito to create a UserPool for an application that does not require any sort of sign in. By setting MFA configuration to 'OPTIONAL', that sort of application would not need to have MFA. 
3. I needed to add some extra logic to handle YAML's reserved keywords `ON` and `OFF` (as well as their variations, i.e `On`, `Off`, etc..). Basically what's happening is that if a user does not wrap either 'ON' or 'OFF' in quotes inside their CloudFormation template, YAML will interpret that as either `true` or `false` and it fail to deploy the stack.
4. This rule and its associated tests basically checks for any variation of the 3 strictly supported keyword values for `MfaConfiguration:` - 'ON'/'OFF'/'OPTIONAL' as well as handles custom parameter JSON files being passed into the template with CFN_NAG.


lib/cfn-nag/custom_rules/CognitoUserPoolMfaConfigurationOnorOptionalRule.rb
```diff
@@ -0,0 +1,61 @@
...
+    ref_value = user_pool.mfaConfiguration.values[0]
+    if user_pool.cfn_model.parameters[ref_value].default.nil?
+    else
+      param_value = user_pool.cfn_model.parameters[ref_value].default
+      violating_key_words.include?(param_value) || yaml_reserved_keyword_boolean_conv?(user_pool, param_value)
+    end
+  end
+
+  def violating_userpool?(user_pool)
+    if user_pool.mfaConfiguration.class == Hash
```
erickascic: can References in cfn-model help with this code?

lib/cfn-nag/custom_rules/CognitoUserPoolMfaConfigurationOnorOptionalRule.rb
```diff
@@ -0,0 +1,61 @@
...
+      violating_userpool?(userpool)
+    end
+
+    violating_userpools.map(&:logical_resource_id)
+  end
+
+  private
+
+  def yaml_reserved_keyword_boolean_conv?(user_pool, param_value = '')
+    if [true, false].include?(user_pool.mfaConfiguration)
```
erickascic: there is a truthy? method in utils somewhere that you should likely use instead

lib/cfn-nag/custom_rules/CognitoUserPoolMfaConfigurationOnorOptionalRule.rb
```diff
@@ -0,0 +1,61 @@
...
+      true
+    elsif [true, false].include?(param_value)
+      true
+    end
+  end
+
+  def keyword_uppercase?(user_pool)
+    violating_key_words = %w[On on OFF Off off Optional optional]
```
erickascic: this code is replicated in a few places.... can't you just do a case insensitive request or force the case for comparison?

lib/cfn-nag/custom_rules/CognitoUserPoolMfaConfigurationOnorOptionalRule.rb
```diff
@@ -0,0 +1,61 @@
...
+require 'cfn-nag/util/enforce_reference_parameter'
+require_relative 'base'
+
+class CognitoUserPoolMfaConfigurationOnorOptionalRule < BaseRule
+  def rule_text
+    "AWS Cognito UserPool should have MfaConfiguration set to 'ON' (MUST be wrapped in quotes) or at least 'OPTIONAL'"
+  end
+
+  def rule_type
+    Violation::FAILING_VIOLATION
```
erickascic: see comment in slack about guess access

lib/cfn-nag/custom_rules/CognitoUserPoolMfaConfigurationOnorOptionalRule.rb
tmcelhattan: Made an update to use the truthy? and not_truthy? methods.

lib/cfn-nag/custom_rules/CognitoUserPoolMfaConfigurationOnorOptionalRule.rb
tmcelhattan: I think it can but I was trying to implement and didn't have lucky getting my tests passing. I pushed an update to help refine this a little but I'm also taking into consideration if there are any default values for params and then if so, check for those ref'd values.

lib/cfn-nag/custom_rules/CognitoUserPoolMfaConfigurationOnorOptionalRule.rb
tmcelhattan: So I was able to eliminate the code duplication for this, but I think I still need to catch every single case of those options (Specifically for the 'ON' and 'OPTIONAL' cases) because CloudFormation will only accept those values in full uppercase.

tmcelhattan:
@erickascic  Did some major refactors to the rule file, also added a util script and a method for checking if a property is referencing a parameter and another method for getting that parameter's Default value.

Issue #68, opened 2017-10-18
Investigate whether AllowUnauthenticatedIdentities  on AWS::Cognito::IdentityPool should ever be false


erickascic:
Make a rule to warn about it being true, or figure out if this should be a failing violation

erickascic:
https://www.pivotaltracker.com/story/show/164842079

PR #372, opened 2020-02-12
Adding warning rule for cognito IdentityPool AllowUnauthenticatedIdentities - #68

tmcelhattan:
#68 
Add rule to warn if AWS::Cognito::IdentityPool has AllowUnauthenticatedIdentities property set to 'true'.  This property should be false but CAN be true only if the proper IAM role and permissions are established for unauthenticated users.


lib/cfn-nag/custom_rules/CognitoIdentityPoolAllowUnauthenticatedIdentitiesRule.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  private
+
+  def violations?(property_value)
+    truthy?(property_value)
+  end
+
+  def violating_identity_pool?(identity_pool)
+    ip_prop = identity_pool.allowUnauthenticatedIdentities
```
erickascic: same issue with parameter subsitution values as for the user pool PR

Issue #10, opened 2016-03-01
Figure out if there any good rules for NACL


erickascic:

erickascic:
https://www.pivotaltracker.com/n/projects/2316748/stories/164842102

PR #373, opened 2020-02-13
#10 Add NACL rules for VPC ID, Protocol, and Port Range

thegonch:
#10 Adds NACL rules around the following criteria:

- For EC2::NetworkAcl, missing VPC ID
- For EC2::NetworkAclEntry, rules around the proper protocols and port ranges to prevent opening traffic on all ports

NOTE: Please feel free to make suggestions around good ideas for NACL rules, how strict we want to be, etc.  Thanks!


lib/cfn-nag/custom_rules/EC2NetworkAclMissingVpcIdRule.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+
+class EC2NetworkAclMissingVpcIdRule < BaseRule
```
erickascic: let's drop this one - if we get better at doing syntax checks, i'd want it in the model, not as a rule

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
```diff
@@ -0,0 +1,49 @@
...
+require 'cfn-nag/violation'
+require_relative 'base'
+
+class EC2NetworkAclEntryPortRangeRule < BaseRule
+  def rule_text
+    'EC2 NetworkACL Entry Port Range must be defined for TCP/UDP protocols and not allow all ports'
+  end
+
+  def rule_type
+    Violation::FAILING_VIOLATION
```
erickascic: i'm thinking this is probably better as a warning?  in my mind, it's legit to develop VPCs that pass the traffic, and security groups bear the brunt.

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  private
+
+  # Port Range is required for protocols "6" (TCP) and "17" (UDP)
+  def tcp_or_udp_protocol?(network_acl_entry)
+    %w[6 17].include?(network_acl_entry.protocol)
+  end
+
+  def port_range_params_exist?(network_acl_entry)
```
erickascic: i think the name is inverted?  this is true if port range _doesn't_ exist

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
```diff
@@ -0,0 +1,49 @@
...
+    network_acl_entry.portRange.nil? ||
+      network_acl_entry.portRange['From'].nil? || network_acl_entry.portRange['To'].nil?
+  end
+
+  def full_port_range?(network_acl_entry)
+    (network_acl_entry.portRange['From'] == '0' && network_acl_entry.portRange['To'] == '65535') ||
+      (network_acl_entry.portRange['From'] == '-1' && network_acl_entry.portRange['To'] == '-1')
```
erickascic: is this an and?  i would confirm by actual converging a template...... might just be either one of from/to being -1 makes it all ports

lib/cfn-nag/custom_rules/EC2NetworkAclEntryProtocolRule.rb
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def icmpv6_protocol?(network_acl_entry)
+    network_acl_entry.protocol == '58' && !network_acl_entry.ipv6CidrBlock.nil? &&
+      !network_acl_entry.icmp.nil? && !network_acl_entry.icmp['Code'].nil? &&
+      !network_acl_entry.icmp['Type'].nil?
+  end
+
+  def violating_network_acl_entries?(network_acl_entry)
+    if tcp_udp_icmp_protocol?(network_acl_entry) ||
```
erickascic: if i'm following this.... i don't think we should restrict to tcp/udp/icmp?  this would make blocking a particular protocol a violation?  at the very least it needs to consider the RuleAction.  https://en.wikipedia.org/wiki/List_of_IP_protocol_numbers

thegonch: So the issue for this comes from here, where using anything that isn't these defined protocols opens up all ports, which I can't imagine we would want: https://docs.aws.amazon.com/AWSEC2/latest/APIReference/API_CreateNetworkAclEntry.html#API_CreateNetworkAclEntry_RequestParameters

> Protocol
The protocol number. A value of "-1" means all protocols. If you specify "-1" or a protocol number other than "6" (TCP), "17" (UDP), or "1" (ICMP), traffic on all ports is allowed, regardless of any ports or ICMP types or codes that you specify. If you specify protocol "58" (ICMPv6) and specify an IPv4 CIDR block, traffic for all ICMP types and codes allowed, regardless of any that you specify. If you specify protocol "58" (ICMPv6) and specify an IPv6 CIDR block, you must specify an ICMP type and code.

Type: String

Required: Yes

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
thegonch: Thanks for pointing this out, helped me catch a glaring error and my fault for not testing it more thoroughly.  So it turns out that you cannot have BOTH the `From` and `To` fields equal the value of '-1', that will fail to build properly as an invalid port range.  You also must have a `From` field so in reality we can ignore the test of `To = '-1'` because that realistically cannot happen (I don't know if we want to concern ourselves with testing if either of these scenarios happens or not).  So the real case here is this, a fully open port range is:
1.  Opening all ports from 0 to 65535
OR
2. Setting `From` to equal '-1' which will set the field in the NACL to be essentially `ALL`

This also continues to be awkward since it states this uses an integer but is always single quoted within Cloudformation templates.

For now, I'm adjusting this match the case I mentioned here.

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
thegonch: You're correct, the name is being adjusted to match that.

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
thegonch: Following up on this and through a lot of testing, I've found that there's no scenario for the NACLEntry PortRange that allows the usage of '-1' as either From or To within Cloudformation, it fails every time.  So I won't be testing for that.

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
thegonch: The only issue that may arise here is that technically the Port Range is required for TCP and UDP protocols.  Should that be a separate rule or not addressed at all?

lib/cfn-nag/custom_rules/EC2NetworkAclEntryProtocolRule.rb
thegonch: Obviously like you said there are cases where allowing all ports to come through is fine, but there are also scenarios here where the Port Range actually gets ignored, so I think we would at least need to warn about that.

lib/cfn-nag/custom_rules/EC2NetworkAclEntryPortRangeRule.rb
```diff
@@ -0,0 +1,48 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+
+class EC2NetworkAclEntryPortRangeRule < BaseRule
+  def rule_text
+    'EC2 NetworkACL Entry Port Range must be defined for TCP/UDP protocols and should probably not allow all ports'
```
erickascic: let's change this verbiage to focus on the port range as opposed to the syntactic rules.  perhaps:

TCP/UDP protocol NetworkACL entries possibly should not allow all ports.

lib/cfn-nag/custom_rules/EC2NetworkAclEntryProtocolRule.rb
```diff
@@ -0,0 +1,58 @@
...
+       icmpv6_protocol?(network_acl_entry)
```
erickascic: 1. do we want to flag protocol == -1 as a warning?
2. wouldn't we want to consider this only when RuleAction == 'allow'?

thegonch: 1.  Are you asking if we should only have `protocol == -1` as a warning or the entire `EC2NetworkAclEntryProtocolRule`?  This rule is currently a failing violation but I can see where it can be a warning as well since it this can be limited elsewhere.  In regards to being explicit about -1, I don't know that it is necessary since we are categorizing just the few protocols instead that would not trigger an opening of all ports.  The test scenario spec/test_templates/yaml/ec2_networkaclentry/ec2_networkaclentry_all_protocols.yml checks for this with a protocol of -1 and triggers the violation as it is now.
2.  The documentation states that anything that isn't one of the specific protocols will be _allowed_ on all ports.  I can play around with testing this a bit more to see how that scenario holds if set it to deny instead.

erickascic: 1. ok makes sense.  -1 is covered alongside anything besides 1, 6, 17.
2. i like warning over fail because there are reason to allow all traffic in conjunction with security groups
3. i believe we only want to warn on Allows that are "open".  if something is set to Deny.... i don't think we warn to warn

Issue #63, opened 2017-10-18
Warn on VPC without a AWS::EC2::FlowLog attached to it


erickascic:
Could be defined in another template so just warn

erickascic:
https://www.pivotaltracker.com/story/show/164842086

twellspring:
Warn if resources_by_type(AWS::EC2::VPC) exists and no resources_by_type(AWS::EC2::FlowLog) exist

Checking more specifically that the VPC is specified in the FlowLog ResourceID is possible but would mean this rule would warn if someone decided to use flow logs on subnets rather than the full vpc.

archetypalsxe:
Not able to assign myself, but I am working on this feature

PR #382, opened 2020-02-23
[#63] Adding a Warning for VPCs Missing FlowLogs

archetypalsxe:


lib/cfn-nag/custom_rules/VpcHasFlowLogRule.rb
```diff
@@ -0,0 +1,27 @@
...
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W58'
+  end
+
+  def audit_impl(cfn_model)
+    violating_policies = cfn_model.resources_by_type('AWS::EC2::VPC')
+                                  .select do
```
erickascic: i'm pretty sure this only works in a coincidental way for the degenerate case.  i believe this will select every vpc if there are no flow logs..... and it will select no vpcs if there is at least one flow log

erickascic: to pick out individual vpcs without flow logs, you'll need to establish the relationship between the model elements and then query against that.

archetypalsxe: @erickascic , I am definitely willing to do that if it's required, but I was going off the comments left by @twellspring on the ticket where he brought up that the flow logs might be attached to subnets rather than the VPC itself. I guess which is the lesser of the two evils?

erickascic: one way or the other, this impl has no traceability back to the attached resource.  as far as subnets go... we could consider them with by synthesizing the model a bit to look for attachments in both places.... on the other hand, i would consider _why_ would the attachment only be at the subnet level?  i have yet to see a security org that singles out particular subnets for compliance, and leaves others out.

Mr-Lizard:
Just a guess, but probably during a debugging / monitoring situation, to
limit extraneous information ?

archetypalsxe:
@erickascic Added additional code for confirming that the flow log references the VPC in question

lib/cfn-nag/custom_rules/VpcHasFlowLogRule.rb
```diff
@@ -0,0 +1,35 @@
...
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W58'
+  end
+
+  def audit_impl(cfn_model)
+    violating_policies = cfn_model.resources_by_type('AWS::EC2::VPC')
```
erickascic: violating_vpcs

lib/cfn-nag/custom_rules/VpcHasFlowLogRule.rb
```diff
@@ -0,0 +1,35 @@
...
+      flowlog_for_vpc(cfn_model, vpc).nil?
+    end
+
+    violating_policies.map(&:logical_resource_id)
+  end
+
+  def flowlog_for_vpc(cfn_model, vpc)
+    cfn_model.resources_by_type('AWS::EC2::FlowLog').find do |flowlog|
+      if flowlog.resourceId && flowlog.resourceId['Ref']
+        flowlog.resourceId['Ref'] == vpc.logical_resource_id
```
erickascic: i think this works.... we could probably tighten this up by building a relationship in cfn-model and we need a more general Ref solution in the model but.... i think we can go with this.  only ask is that you add a test case for > 1 VPC in the sample template

archetypalsxe:
@erickascic Added additional tests and correcting verbiage accordingly

PR #385, opened 2020-02-24
init commit for issue 72 - API Gateway AuthType property - warning fo…

tmcelhattan:
Set a WARNING rule for AWS::ApiGateway::Method resource if `AuthorizationType` property is set to `NONE`. This property should be set to either of the following: `AWS_IAM`, `CUSTOM`, `COGNITO_USER_POOLS`.

### DIFF
--- a/lib/cfn-nag/custom_rules/ApiGatewayMethodAuthorizationTypeRule.rb
+++ b/lib/cfn-nag/custom_rules/ApiGatewayMethodAuthorizationTypeRule.rb
@@ -18,7 +18,7 @@ class ApiGatewayMethodAuthorizationTypeRule < BaseRule
 
   def audit_impl(cfn_model)
     violating_deployments = cfn_model.resources_by_type('AWS::ApiGateway::Method').select do |method|
-      method.authorizationType.nil? || method.authorizationType.to_s.downcase == 'none'
+      method.authorizationType.to_s == '' || method.authorizationType.to_s.casecmp('none').zero?
     end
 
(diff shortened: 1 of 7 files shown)


lib/cfn-nag/custom_rules/ApiGatewayMethodAuthorizationTypeRule.rb
```diff
@@ -0,0 +1,26 @@
...
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W58'
+  end
+
+  def audit_impl(cfn_model)
+    violating_deployments = cfn_model.resources_by_type('AWS::ApiGateway::Method').select do |method|
+      method.authorizationType.nil? || method.authorizationType.to_s.downcase == 'none'
```
erickascic: technically nil.to_s == ''

erickascic: to_s.downcase is cool, but I believe we've got a bunch of prior art/style around using strcmpi

tmcelhattan: @erickascic  Just pushed up some slight changes. Is that what you had in mind?

erickascic: ok i made a half stupid comment.  i like your original .nil? call better.  i was thinking nil being empty would fall through properly in the casecmp but it won't.  sorry for the runaround.

PR #393, opened 2020-03-02
#59 Two rules to validate EMR SecurityConfigurations

### LINKED ISSUE #59: Investigate AWS::EMR::SecurityConfiguration
Looks like there is some potential to check the values here, but requires some research since the format isn't specific by cloudformation

pshelby:
Closes #59 

### Description

As a first swag at rules for EMR, I created one rule to ensure the `SecurityConfiguration` property exists on each Cluster and another rule to validate SecurityConfiguration resources for enabled `at rest` and `in transit` encryptions and necessary accompanying parameters.

### Testing

1. Created new rpsec tests for EMR Cluster and SecurityConfiguration rules.
2. All rubocop and rspec tests passed.


lib/cfn-nag/custom_rules/EMRClusterSecurityConfigurationAttachedRule.rb
```diff
@@ -0,0 +1,31 @@
...
+  end
+
+  def rule_id
+    'W60'
+  end
+
+  def audit_impl(cfn_model)
+    violating_emr_clusters = cfn_model.resources_by_type('AWS::EMR::Cluster').select do |cluster|
+      # Warn if SecurityConfiguration property is not set...
+      next cluster unless cluster.securityConfiguration
```
erickascic: this works - one style suggestion/nit is that "next" is sort of an imperative way to look at this loop vs. a predicate.... you could rewrite this bit to  be: cluster.securityConfiguration.nil? || security_config_reference_not_found?(cluster)

lib/cfn-nag/custom_rules/EMRSecurityConfigurationEncryptionsEnabledAndConfiguredRule.rb
```diff
@@ -0,0 +1,65 @@
...
+  def bad_security_config?(security_config_object)
+    # Poorly formatted SecurityConfiguration
+    return true unless security_config_object.securityConfiguration['EncryptionConfiguration']
+
+    encryption_config = security_config_object.securityConfiguration['EncryptionConfiguration']
+
+    # Either encryption type disabled
+    return true unless encryption_config['EnableAtRestEncryption'] && encryption_config['EnableInTransitEncryption']
+
+    bad_at_rest_encryption?(encryption_config)
```
erickascic: do we need a disjunction here... the return value for this will be lost in the ether

lib/cfn-nag/custom_rules/EMRClusterSecurityConfigurationAttachedRule.rb
pshelby: Thanks!  Updated to make this cleaner in the latest commit.

lib/cfn-nag/custom_rules/EMRSecurityConfigurationEncryptionsEnabledAndConfiguredRule.rb
pshelby: Ah, good call.  My brain thought that was a return, but the boolean would get lost.  I reworked this, the JSON parsing logic, and the rspec tests to return correct values for all incorrect scenarios.

PR #412, opened 2020-03-18
Fix Omissions For W58

zf-jmac:
**Issue**
The existing conditions for W58 - _Lambda functions require permission to write CloudWatch Logs_ omits some alternative managed policies which may be utilised by the function that do provide CloudWatch Logs permissions. 

This leads to a situation where false positive warnings are generated, where a function is flagged as not having write access to CloudWatch Logs when it actually does.

**Fix**
A number of additional managed policies designed for usage with Lambda that do provide CloudWatch Logs write permissions have been added to reduce false positives.


erickascic:
@pshelby looks good to me, but can you give this one a gander?

PR #417, opened 2020-03-24
#378 Created missing_property_base_rule and applied to SageMaker EndpointConfig/NotebookInstance

### LINKED ISSUE #378: SageMaker KMS keys
Warn on no kms keys for:

AWS::SageMaker::EndpointConfig
AWS::SageMaker::NotebookInstance

we should have a few existing rules like this and possibly a base class to derive from?

pshelby:
Closes #378 

### Description

In order to detect missing properties, I created a base rule which will be flexible enough to detect any missing property for any type of resource.  I built the rule on top of and reworked the existing `password_rule_spec_helper` logic so it is more generic and can be extended as well.

1. Reworked `password_rule_spec_helper` to `custom_rule_spec_helper` so it can be used in more generic cases.
2. Added a missing_property_base_rule to detect missing properties in given resources.
3. Created custom rules to warn on missing KmsKeyId properties for both SageMaker EndpointConfig and NotebookInstance resources.
4. Moved "rule test sets" to their own file for ease of reading and extension.
5. Wrote appropriate tests.

### Testing

1. Created test templates for missing and existing properties.
2. All rubocop and rspec tests pass.

### DIFF
--- a/lib/cfn-nag/custom_rules/SageMakerEndpointConfigKmsKeyIdRule.rb
+++ b/lib/cfn-nag/custom_rules/SageMakerEndpointConfigKmsKeyIdRule.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'missing_property_base_rule'
+
+class SageMakerEndpointConfigKmsKeyIdRule < MissingPropertyBaseRule
+  def rule_text
+    'SageMaker EndpointConfig must have a KmsKeyId property set.'
+  end
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W1200'
+  end
+
+  def resource_type
+    'AWS::SageMaker::EndpointConfig'
+  end
+
+  def property_name
+    :kmsKeyId
+  end
+end
--- a/spec/custom_rules/AmazonMQBrokerEncryptOptionsRule_spec.rb
+++ b/spec/custom_rules/AmazonMQBrokerEncryptOptionsRule_spec.rb
@@ -1,5 +1,6 @@
 require 'spec_helper'
-require 'password_rule_spec_helper'
+require 'custom_rule_spec_helper'
+require 'custom_rule_test_sets'
 require 'cfn-model'
 require 'cfn-nag/custom_rules/AmazonMQBrokerEncryptionOptionsRule'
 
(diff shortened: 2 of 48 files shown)


lib/cfn-nag/custom_rules/SageMakerEndpointConfigKmsKeyIdRule.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'missing_property_base_rule'
+
+class SageMakerEndpointConfigKmsKeyIdRule < MissingPropertyBaseRule
+  def rule_text
+    'SageMaker EndpointConfig must have a KmsKeyId property set.'
```
erickascic: not sure we've ever been consistent about this, but "must" sounds a fail vs. a warning.  maybe "should" instead?

spec/custom_rules/AmazonMQBrokerEncryptOptionsRule_spec.rb
```diff
@@ -1,5 +1,6 @@
 require 'spec_helper'
-require 'password_rule_spec_helper'
+require 'custom_rule_spec_helper'
```
erickascic: we could probably jam all of this stuff into spec_helper going forward? (to avoid 45 deltas :) )

lib/cfn-nag/custom_rules/SageMakerEndpointConfigKmsKeyIdRule.rb
pshelby: Good point!  I'll update to "should" in the next commit.

spec/custom_rules/AmazonMQBrokerEncryptOptionsRule_spec.rb
pshelby: True, I considered that, however all the functions in `custom_rule_spec_helper` are more geared toward the types of rules that @phelewski created in his password rule framework.  I just found the [boolean_base_rule.rb](lib/cfn-nag/custom_rules/boolean_base_rule.rb) however, and that may fulfill my needs for the "missing property base rule".  I'll do more research and possibly remove most of this PR.  😄

pshelby:
Closing for further research and changes.  Will reopen in the future.

Issue #409, opened 2020-03-17
Rule for considering order of NACL entries


erickascic:
Ineffective or misconfigured DENY rules promotes ‘overly-permissive’ access to a VPC. This results in attacks, such as DoS or DDoS. Be mindful of the order of the DENY rules within your Network ACLs. This is crucial, as ACLs are evaluated in order. For example, in the below image, the DENY rule is defined to block inbound traffic to vulnerable port 2049. If the rule does not block access to everyone (0.0.0.0/0), the inbound DENY rule is declared ineffective and should be reconfigured to protect against attacks

basically, attempt to eval or partial eval on the rules to check whether some of the rules are ineffective because of earlier rules in the template.

PR #423, opened 2020-04-01
#409 Ineffective NACL rules

thegonch:
In reference to stelligent/cfn_nag#409.

- Updates cfn-model gem to 0.4.28
- Creates rules around ineffective NACL entry rules, which are separated by egress and ingress since they can each have the same reused rule number, including:
  - Failing on duplicate rule numbers
  - Warning on deny rules that do not cover the full CIDR range
  - Warning when reusing port ranges
- Spec tests for the above


lib/cfn-nag/custom_rules/EC2NetworkAclEntryDuplicateRule.rb
```diff
@@ -0,0 +1,47 @@
...
+  end
+
+  def rule_id
+    'F79'
+  end
+
+  def audit_impl(cfn_model)
+    nacl_violating_egress_entries = []
+    nacl_violating_ingress_entries = []
+    cfn_model.resources_by_type('AWS::EC2::NetworkAcl').select do |nacl|
```
erickascic, line 22: do you mean each here instead of select?

lib/cfn-nag/custom_rules/EC2NetworkAclEntryDuplicateRule.rb
```diff
@@ -0,0 +1,47 @@
...
+    nacl_violating_egress_entries.map(&:logical_resource_id) + nacl_violating_ingress_entries.map(&:logical_resource_id)
+  end
+
+  private
+
+  def duplicate_rule_numbers?(cfn_model, nacl_entries)
+    nacl_entry_resources = []
+    rule_numbers = []
+    duplicates = []
+    nacl_entries.select do |nacl_entry|
```
erickascic, line 36: i'm a bit confused.  are these select supposed to be each?  you could probably rewrite this code a lot tighter if you used select to actually select instead of using the predicate as a side effect to update the "outside" arrays

lib/cfn-nag/custom_rules/EC2NetworkAclEntryIneffectiveDenyRule.rb
```diff
@@ -0,0 +1,46 @@
...
+
+    violating_egress_entries.map(&:logical_resource_id) + violating_ingress_entries.map(&:logical_resource_id)
+  end
+
+  private
+
+  def deny_does_not_cover_all_cidrs?(cfn_model, nacl_entries)
+    nacl_entry_resources = []
+    nacl_entries.select do |nacl_entry|
+      nacl_entry_resources << cfn_model.resource_by_id(nacl_entry)
```
erickascic, line 38: i would have expect the Nacl in cfn-model to have a collection of NaclEntry objects as opposed to a collection of logical resource ids (so the rule writer doesn't have to dereference them)

erickascic, line 38: beating the dead horse here but you could do a map here and assign it to nacl_entry_resources instead of this side effect approach

lib/cfn-nag/custom_rules/EC2NetworkAclEntryReusedPortsRule.rb
```diff
@@ -0,0 +1,48 @@
...
+
+  def reused_ports?(cfn_model, nacl_entries)
+    nacl_entry_resources = []
+    ports = []
+    reused = []
+    nacl_entries.select do |nacl_entry|
+      nacl_entry_resources << cfn_model.resource_by_id(nacl_entry)
+    end
+    nacl_entry_resources.select do |nacl_entry_resource|
+      if ports.include?(nacl_entry_resource.portRange)
```
erickascic, line 41: can you check overlap?

lib/cfn-nag/custom_rules/EC2NetworkAclEntryIneffectiveDenyRule.rb
```diff
@@ -0,0 +1,46 @@
...
+    end
+    nacl_entry_resources.select do |nacl_entry_resource|
+      nacl_entry_resource.ruleAction == 'deny' && ((!nacl_entry_resource.cidrBlock.nil? &&
```
erickascic, line 41: consider whether this expression could be named and decomposed into a predicate function

lib/cfn-nag/custom_rules/EC2NetworkAclEntryIneffectiveDenyRule.rb
thegonch: Ok, I've been going through this for a little bit today.  It was based on the API gateway work but I agree it's a bit tedious this way.  Would this necessitate creating a separate NetworkAclEntry parser?  

In testing now with cfn-model, here's an example NetworkAclEntry resource item as it stands:

`[#<AWS::EC2::NetworkAclEntry:0x0000556bb0856e40 @cfn_model=#<CfnModel:0x0000556bb0861980 @parameters={}, @resources={"testvpc1"=>#<AWS::EC2::VPC:0x0000556bb08611b0 @cfn_model=#<CfnModel:0x0000556bb0861980 ...>, @logical_resource_id="testvpc1", @resource_type="AWS::EC2::VPC", @metadata=nil, @cidrBlock="10.0.0.0/16", @enableDnsSupport=true, @enableDnsHostnames=true, @instanceTenancy="default">, "myNetworkAcl"=>#<AWS::EC2::NetworkAcl:0x0000556bb0857430 @cfn_model=#<CfnModel:0x0000556bb0861980 ...>, @network_acl_egress_entries=[], @network_acl_ingress_entries=[], @resource_type="AWS::EC2::NetworkAcl", @vpcId="testvpc1", @logical_resource_id="myNetworkAcl", @metadata=nil>, "EgressEntry1"=>#<AWS::EC2::NetworkAclEntry:0x0000556bb0856e40 ...>}, @conditions={}, @globals={}, @raw_model={"Resources"=>{"testvpc1"=>{"Type"=>"AWS::EC2::VPC", "Properties"=>{"CidrBlock"=>"10.0.0.0/16", "EnableDnsSupport"=>true, "EnableDnsHostnames"=>true, "InstanceTenancy"=>"default"}}, "myNetworkAcl"=>{"Type"=>"AWS::EC2::NetworkAcl", "Properties"=>{"VpcId"=>"testvpc1"}}, "EgressEntry1"=>{"Type"=>"AWS::EC2::NetworkAclEntry", "Properties"=>{"NetworkAclId"=>{"Ref"=>"myNetworkAcl"}, "Protocol"=>"6", "RuleAction"=>"allow", "RuleNumber"=>"100", "CidrBlock"=>"10.0.0.0/16", "Egress"=>true, "PortRange"=>{"From"=>"443", "To"=>"443"}}}}}, @line_numbers={}>, @logical_resource_id="EgressEntry1", @resource_type="AWS::EC2::NetworkAclEntry", @metadata=nil, @networkAclId={"Ref"=>"myNetworkAcl"}, @protocol="6", @ruleAction="allow", @ruleNumber="100", @cidrBlock="10.0.0.0/16", @egress=true, @portRange={"From"=>"443", "To"=>"443"}>]
`
This is including more than is desired.

erickascic: yea gw was a tough one.... the relationships among the model elements (on the aws side) don't make obvious sense and some are implicit.  a good rule though is to build any "referencing" or synthesizing into the underlying cfn-model so that rule developer can just manipulate objects/properties

thegonch: Do you feel it still helps to create separate ingress and egress objects for NACL since they seem to generally have separate rule applications?  Or does it make sense to offer a third option that includes all of them?

erickascic: i think you can have a single Entry object that covers both of them.  given it's an object you can hide some of the "statefulness" behind human readable predicates too e.g. is_egress?

thegonch:
> there a number of places in this code where select/each/map are/aren't being used the way they should be. i'd advocate using select/map where you can instead of each with outside state (the code is using select with outside state which doesn't make sense). i think the rules themselves make sense. if the port range stuff overlaps... think it's a good way to warn developers they could be making trouble for themselves.

@erickascic I've updated the code here though it is pending the review/merge of https://github.com/stelligent/cfn-model/pull/77 to pull in the updated cfn-model gem.

lib/cfn-nag/custom_rules/EC2NetworkAclEntryIneffectiveDenyRule.rb
```diff
@@ -14,7 +14,7 @@ def rule_type
   end
 
   def rule_id
-    'W69'
+    'W70'
```
erickascic, line 17: pull from master.  just merged W70 from an external contributor

lib/cfn-nag/custom_rules/EC2NetworkAclEntryDuplicateRule.rb
thegonch: Updated

lib/cfn-nag/custom_rules/EC2NetworkAclEntryIneffectiveDenyRule.rb
thegonch: Updated

thegonch: Updated

lib/cfn-nag/custom_rules/EC2NetworkAclEntryReusedPortsRule.rb
thegonch: Added, good idea

lib/cfn-nag/custom_rules/EC2NetworkAclEntryDuplicateRule.rb
thegonch: Updated

lib/cfn-nag/custom_rules/EC2NetworkAclEntryDuplicateRule.rb
erickascic: i think i get this logic but it took me a while of staring at the conjunction with the next and wondering about operator precedence.

the first thing i'd do is drop the ? for methods that are returning non-boolean values....

would this "one-liner" solve the problem?

nacl_entries.group_by { |entry| entry.ruleNumber }.select { |_, entries| entries.size > 1 }.map { |_, entries| entries }.flatten

thegonch: Updates are made, please re-review @erickascic when you get a chance.  Thanks!

lib/cfn-nag/custom_rules/EC2NetworkAclEntryReusedPortsRule.rb
```diff
@@ -0,0 +1,81 @@
...
+        nacl_entry
+      end
+    end
+  end
+
+  def update_overlapping_port_range(nacl_entry_range, ports)
+    if ports.nil?
+      nacl_entry_range
+    else
+      port_min = nacl_entry_range.min < ports.min ? nacl_entry_range.min : ports.min
```
erickascic, line 51: i believe this code is same as [nacl_entry_rnage.min, ports.min].min

erickascic: not a big deal - but you save 1 point per statement on CCM

lib/cfn-nag/custom_rules/EC2NetworkAclEntryReusedPortsRule.rb
```diff
@@ -0,0 +1,81 @@
...
+    'W72'
+  end
+
+  def audit_impl(cfn_model)
+    violating_nacl_entries = []
+    cfn_model.resources_by_type('AWS::EC2::NetworkAcl').each do |nacl|
+      violating_nacl_entries += violating_nacl_entries(nacl)
+    end
+
+    violating_nacl_entries.uniq.map(&:logical_resource_id)
```
erickascic, line 26: is the use of uniq necessary and tested?  i don't believe hash or eql? are defined for the nacl entry model element?

Issue #384, opened 2020-02-24
Cloudfront TLS-1.2 rule


pethers:
missing rule for checking MinimumProtocolVersion is set to TLS 1.2 from https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-viewercertificate.html

Allowed Values: SSLv3 | TLSv1 | TLSv1.1_2016 | TLSv1.2_2018 | TLSv1_2016

fail on SSLv3 | TLSv1 | TLSv1.1_2016 | TLSv1_2016

PR #426, opened 2020-04-04
https://github.com/stelligent/cfn_nag/issues/384 Cloudfront TLS 1.2 rule

pethers:
When using CloudFrontDefaultCertificate  (you set CloudFrontDefaultCertificate to true), CloudFront automatically sets the security policy to TLSv1 regardless of the value that you set here.)

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-viewercertificate.html

### DIFF
--- a/lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
+++ b/lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+
+class CloudfrontMinimumProtocolVersionRule < BaseRule
+  def rule_text
+    'Cloudfront should use minimum protocol version TLS 1.2'
+  end
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W70'
+  end
+
+  def audit_impl(cfn_model)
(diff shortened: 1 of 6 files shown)


lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  def rule_id
+    'W70'
+  end
+
+  def audit_impl(cfn_model)
+    violating_distributions = cfn_model.resources_by_type('AWS::CloudFront::Distribution')
+                                       .select do |distribution|
+      distribution.distributionConfig['ViewerCertificate'].nil? || distribution.distributionConfig['ViewerCertificate']['MinimumProtocolVersion'].nil? || distribution.distributionConfig['ViewerCertificate']['MinimumProtocolVersion'] != 'TLSv1.2_2018' || !distribution.distributionConfig['ViewerCertificate']['CloudFrontDefaultCertificate'].nil? || distribution.distributionConfig['ViewerCertificate']['CloudFrontDefaultCertificate']
```
erickascic, line 22: this line won't pass the rubocop checks for number of characters.  can you break this down into smaller predicate functions and then bubble them up here?

pethers:
Thanks for the feedback, will fix shortly.

PR #436, opened 2020-04-27
Add rule that checks if DynamoDB table explicitly indicates billing m…

dscpinheiro:
…ode (provisioned or pay per request)

### DIFF
--- a/lib/cfn-nag/custom_rules/DynamoDBBillingModeRule.rb
+++ b/lib/cfn-nag/custom_rules/DynamoDBBillingModeRule.rb
@@ -1,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require_relative 'base'
+
+class DynamoDBBillingModeRule < BaseRule
+  def rule_text
+    'DynamoDB table should have billing mode set to either PAY_PER_REQUEST or PROVISIONED'
+  end
+
+  def rule_type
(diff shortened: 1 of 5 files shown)


lib/cfn-nag/custom_rules/DynamoDBBillingModeRule.rb
```diff
@@ -0,0 +1,26 @@
...
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W73'
+  end
+
+  def audit_impl(cfn_model)
+    violating_ddb_tables = cfn_model.resources_by_type('AWS::DynamoDB::Table').select do |table|
+      table.billingMode.nil? || (table.billingMode != 'PAY_PER_REQUEST' && table.billingMode != 'PROVISIONED')
```
erickascic, line 21: if i'm reading the docs correctly - these are the only two legal values right now?  if so, i'd only do the nil check in case new values are introduced later. i'm presuming the point here is just to make sure the developer considers the billing mode and chooses one as opposed to picking a particular one?

dscpinheiro: > make sure the developer considers the billing mode and chooses one as opposed to picking a particular one?

Yes, that's pretty much it. Our thinking is that it's better to pick one explicitly than just use the default and then run into throttling issues later.

PR #437, opened 2020-04-27
Add rule that checks if DynamoDB table enables server-side encryption…

dscpinheiro:
… using a CMK stored in KMS

### DIFF
--- a/lib/cfn-nag/custom_rules/DynamoDBEncryptionRule.rb
+++ b/lib/cfn-nag/custom_rules/DynamoDBEncryptionRule.rb
@@ -1,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'cfn-nag/violation'
+require 'cfn-nag/util/truthy'
+require_relative 'base'
+
+class DynamoDBEncryptionRule < BaseRule
+  def rule_text
+    'DynamoDB table should have encryption enabled using a CMK stored in KMS'
+  end
+
+  def rule_type
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W74'
+  end
+
+  def audit_impl(cfn_model)
+    violating_ddb_tables = cfn_model.resources_by_type('AWS::DynamoDB::Table').select do |table|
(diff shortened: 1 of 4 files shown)


dscpinheiro:
Did you have a chance to look at this PR @erickascic?

lib/cfn-nag/custom_rules/DynamoDBEncryptionRule.rb
```diff
@@ -0,0 +1,27 @@
...
+    Violation::WARNING
+  end
+
+  def rule_id
+    'W74'
+  end
+
+  def audit_impl(cfn_model)
+    violating_ddb_tables = cfn_model.resources_by_type('AWS::DynamoDB::Table').select do |table|
+      table.sSESpecification.nil? || !truthy?(table.sSESpecification['SSEEnabled'].to_s)
```
erickascic, line 22: the point of this rule is only to flag an AWS owned key v. AWS managed key? don't care about the KmsKeyId property?

dscpinheiro: Yes, it's just to check that the key is not owned by Amazon DynamoDB.

erickascic: philosophical question for you - do you think not specifying explicit customer CMK for resources should be a warning?  we have a few resources where it does warn... but there are about 40 other resources we don't check.  I was thinking to handle this consistently and make it a warning

dscpinheiro: I talked to a co-worker and we agreed it probably shouldn't, as it depends a lot on whatever the customer needs are. For some, just using an AWS CMK is enough, but others might need to use a customer CMK for compliance reasons.

dscpinheiro: I talked to a co-worker as well and we agreed it probably shouldn't be a warning, as it depends a lot on whatever the customer needs are. For some, just using an AWS CMK is enough, but others might need to use a customer CMK for further compliance reasons.

PR #438, opened 2020-04-27
Add rule that checks if RDS instance has a backup retention period gr…

dscpinheiro:
…eater than 0

### DIFF
--- a/lib/cfn-nag/custom_rules/RDSInstanceBackupRetentionPeriodRule.rb
+++ b/lib/cfn-nag/custom_rules/RDSInstanceBackupRetentionPeriodRule.rb
@@ -20,9 +20,23 @@ class RDSInstanceBackupRetentionPeriodRule < BaseRule
     rds_dbinstances = cfn_model.resources_by_type('AWS::RDS::DBInstance')
 
     violating_rdsinstances = rds_dbinstances.select do |instance|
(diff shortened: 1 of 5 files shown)


lib/cfn-nag/custom_rules/RDSInstanceBackupRetentionPeriodRule.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def rule_id
+    'W75'
+  end
+
+  def audit_impl(cfn_model)
+    rds_dbinstances = cfn_model.resources_by_type('AWS::RDS::DBInstance')
+
+    violating_rdsinstances = rds_dbinstances.select do |instance|
+      instance.backupRetentionPeriod && instance.backupRetentionPeriod == 0
```
erickascic, line 23: even though the cfn spec says this property is an int.... i've seen cfn exhibit "weak typing" and accept string versions of numbers.  so in this case, the string "0" wouldn't be a violation.  can you confirm the behavior or do a to_i/to_s to normalize?  thx!

dscpinheiro: Fixed in https://github.com/stelligent/cfn_nag/pull/438/commits/2973208155f28e94cd777faa571ba1beb899a417

PR #439, opened 2020-04-27
Add rule that checks if RDS instance has deletion protection enabled

dscpinheiro:


spec/test_templates/json/rds_instance/rds_instance_no_echo_with_default_password.json
```diff
@@ -14,7 +14,8 @@
         "MasterUserPassword": {
           "Ref": "Password"
         },
-        "StorageEncrypted": "true"
+        "StorageEncrypted": "true",
+        "DeletionProtection": true
```
dscpinheiro, line 18: As the new rule is failing violation, I had to update some existing test templates.

spec/cfn_nag_integration/cfn_nag_rds_instance_spec.rb
```diff
@@ -81,13 +81,13 @@
                 id: 'F23', type: Violation::FAILING_VIOLATION,
                 message: 'RDS instance master user password must not be a plaintext string or a Ref to a NoEcho Parameter with a Default value.',
                 logical_resource_ids: %w[BadDb1 BadDb2],
-                line_numbers: [14, 28]
+                line_numbers: [14, 29]
```
dscpinheiro, line 84: In this test file, I only updated the line numbers. Please let me know if I should check for the new violation as well.

Issue #448, opened 2020-05-01
W66 incorrectly triggered


candrews:
Given this template:
```
---
AWSTemplateFormatVersion: '2010-09-09'
Description: Test
Resources:
  PublicNetworkAcl:
    Type: AWS::EC2::NetworkAcl
    Properties:
      Tags:
        - Key: Network
          Value: Public
  InboundHTTPPublicNetworkAclEntry:
    Type: AWS::EC2::NetworkAclEntry
    Properties:
      NetworkAclId: !Ref 'PublicNetworkAcl'
      RuleNumber: 100
      Protocol: 6
      RuleAction: allow
      Egress: false
      CidrBlock: '0.0.0.0/0'
      PortRange:
        From: 80
        To: 80
```

I get this warning (using cfn_nag 0.5.54):
```
| WARN W66
|
| Resources: ["InboundHTTPPublicNetworkAclEntry"]
| Line Numbers: [12]
|
| To avoid opening all ports for Allow rules, EC2 NetworkACL Entry Protocol should be either 6 (for TCP), 17 (for UDP), 1 (for ICMP), or 58 (for ICMPv6, which must include an IPv6 CIDR block, ICMP type, and code).
```

Protocol 6 is being used so the warning is clearly incorrect.

erickascic:
fix coming out shortly.  bad comparison between int and string for '6'/6

erickascic:
@candrews pls give 0.5.57 a go

candrews:
> pls give 0.5.57 a go

Success!

0.5.56 reports a warning, 0.5.57 does not :tada:

PR #451, opened 2020-05-01
#448 normalize NACL attribute comparisons between string/int

erickascic:
For example, protocol '6' and 6 should match

### DIFF
--- a/lib/cfn-nag/ip_addr.rb
+++ b/lib/cfn-nag/ip_addr.rb
@@ -8,7 +8,7 @@ module IpAddr
   end
 
   def ip6_localhost?(egress)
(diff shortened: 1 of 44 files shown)


lib/cfn-nag/ip_addr.rb
```diff
@@ -8,7 +8,7 @@ def ip4_localhost?(egress)
   end
 
   def ip6_localhost?(egress)
-    egress.cidrIpv6.is_a?(String) && egress.cidrIpv6 == '::1/128'
+    egress.cidrIpv6.is_a?(String) && (egress.cidrIpv6 == '::1/128' || egress.cidrIpv6 == ':1/128')
```
thegonch, line 11: Only other thing I would consider would be a test for this particular scenario of single vs. double colon.

PR #470, opened 2020-05-26
Update accessLogSettings key for API GatewayV2

shadycuz:
A bug existed in cfn_nag where it would flag valid templates that had accessLogSettings defined as not being valid. It appears that they changed the key from v1 to v2. 

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigateway-stage.html

I was unable to run the tests locally as I can't install docker in wsl 😊 , there may need to be more changes than this but this was the obvious ones to me.


pethers:
Think you are right https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-apigatewayv2-stage.html uses AccessLogSettings. Thanks for fixing this.

erickascic:
released in 0.6.4

PR #469, opened 2020-05-22
Adding Pre commit config to allow for integration to the pre commit…

JBarnes94:
… framework

PR #482, opened 2020-08-20
Allow running cfn_nag as a pre-commit hook

ikben:
- This will make it easy for users to run cfn_nag before committing any file to git.
- See https://pre-commit.com/


ikben:
This was quick to write, so I didn't check if there was already #469  for this.

PR #489, opened 2020-10-01
Relax AWS S3 constraints

arothian:
Allow consumers of cfn_nag to update to more recent versions of `aws-sdk-s3` while still not allowing major version bumps


Gemfile.lock
```diff
@@ -2,7 +2,7 @@ PATH
   remote: .
   specs:
     cfn-nag (0.0.0)
-      aws-sdk-s3 (~> 1.60.1)
+      aws-sdk-s3 (~> 1.76)
```
thegonch, line 5: Was this version chosen specifically to avoid the bump in the aws-sdk-core minimum version?

arothian: This version was specifically choosen as it includes some fixes in the s3 client around encryption - https://github.com/aws/aws-sdk-ruby/blob/master/gems/aws-sdk-s3/CHANGELOG.md#1760-2020-08-07

These don't directly impacted `cfn_nag`, but consumers of `cfn_nag` will now be able to update to newer versions of this gem within their Gemfile.

thegonch (approved):

PR #496, opened 2020-11-06
Update MinimumProtocolVersion to the most recent TLSv1.2_2019

osterik:
tiny fix for waring `(W70) Cloudfront should use minimum protocol version TLS 1.2`
when  
`  MinimumProtocolVersion: TLSv1.2_2019`

proof: [AWS::CloudFront::Distribution ViewerCertificate](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-cloudfront-distribution-viewercertificate.html#cfn-cloudfront-distribution-viewercertificate-minimumprotocolversion)

### DIFF
--- a/lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
+++ b/lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
@@ -32,7 +32,7 @@ class CloudfrontMinimumProtocolVersionRule < BaseRule
   end
 
   def cert_has_bad_tls_version?(viewer_certificate)
-    viewer_certificate['MinimumProtocolVersion'].nil? || viewer_certificate['MinimumProtocolVersion'] != 'TLSv1.2_2018'
+    viewer_certificate['MinimumProtocolVersion'].nil? || viewer_certificate['MinimumProtocolVersion'] != 'TLSv1.2_2019'
   end
 
   def override_tls_config?(viewer_certificate)
--- a/lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
+++ b/lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
@@ -5,7 +5,7 @@ require_relative 'base'
 
 class CloudfrontMinimumProtocolVersionRule < BaseRule
   def rule_text
-    'Cloudfront should use minimum protocol version TLS 1.2'
+    'Cloudfront should use the most advanced available TLS protocol version'
   end
 
   def rule_type


pethers:
Thanks for creating this pull request, the rule "Cloudfront should use minimum protocol version TLS 1.2"  so needs to accept both 'TLSv1.2_2018' and TLSv1.2_2019'  or deny all policies that allow below TLS1.2.

osterik:
Maybe it would be better to re-name this rule to "_Cloudfront should use the most advanced TLS protocol version_"?

Nowadays, even with `MinimumProtocolVersion=TLSv1.2_2019`, check on https://www.ssllabs.com/ssltest/analyze.html page says that 2 cipher suites are already weak (TLS_ECDHE_RSA_WITH_AES_256_CBC_SHA384 & 
TLS_ECDHE_RSA_WITH_AES_128_CBC_SHA256 )

osterik:
@pethers , `rule_text` changed to 'Cloudfront should use the most advanced available TLS protocol version'

pethers:
@thegonch can you please review this ?

pethers:
Can anyone at Stelligent please review this ?

lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
```diff
@@ -5,7 +5,7 @@
 
 class CloudfrontMinimumProtocolVersionRule < BaseRule
   def rule_text
-    'Cloudfront should use minimum protocol version TLS 1.2'
+    'Cloudfront should use the most advanced available TLS protocol version'
```
arothian, line 8: I think the intent here is that we enforce a secure choice for this parameter, rather than simply "the latest and greatest". Maybe rewording this a bit makes sense.
```suggestion
    'Cloudfront should use a secure minimal protocol version'
```

arothian (commented):
Minor suggestion on renaming the rule text.

I think the tests in `CloudfrontMinimumProtocolVersionRule_spec.rb` need to be adjusted to handle the new case.

lib/cfn-nag/custom_rules/CloudfrontMinimumProtocolVersionRule.rb
thegonch: Note that this will also require updating of the spec test language here: https://github.com/stelligent/cfn_nag/blob/master/spec/cfn_nag_integration/cfn_nag_cloudfront_distribution_spec.rb#L29

thegonch (changes requested):
Aside from the other comment I made about the spec test update for the warning language, this change in general requires updating the following spec test and its test template:

https://github.com/stelligent/cfn_nag/blob/master/spec/custom_rules/CloudfrontMinimumProtocolVersionRule_spec.rb#L30

https://github.com/stelligent/cfn_nag/blob/master/spec/test_templates/json/cloudfront_distribution/cloudfront_with_tls_12.json#L8
https://github.com/stelligent/cfn_nag/blob/master/spec/test_templates/json/cloudfront_distribution/cloudfront_with_tls_12.json#L23

arothian:
Closing as #531 addresses this.

Issue #480, opened 2020-07-30
documented cfn_nag_rules


PatMyron:
It'd be nice to see all the current rules without needing to install, update, and run `cfn_nag_rules`, and it'd be nice to see the rule history over time, similar to https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md#rules-1

A regularly updated file with the current plaintext output of `cfn_nag_rules` should be good enough:
```
WARNING VIOLATIONS:
W1 Specifying credentials in the template itself is probably not the safest thing
W2 Security Groups found with cidr open to world on ingress.  This should never be true on instance.  Permissible on ELB
W5 Security Groups found with cidr open to world on egress
W9 Security Groups found with ingress cidr that is not /32
W10 CloudFront Distribution should enable access logging
W11 IAM role should not allow * resource on its permissions policy
W12 IAM policy should not allow * resource
W13 IAM managed policy should not allow * resource
W14 IAM role should not allow Allow+NotAction on trust permissions
W15 IAM role should not allow Allow+NotAction
W16 IAM policy should not allow Allow+NotAction
W17 IAM managed policy should not allow Allow+NotAction
W18 SQS Queue policy should not allow Allow+NotAction
W19 SNS Topic policy should not allow Allow+NotAction
W20 S3 Bucket policy should not allow Allow+NotAction
W21 IAM role should not allow Allow+NotResource
W22 IAM policy should not allow Allow+NotResource
W23 IAM managed policy should not allow Allow+NotResource
W24 Lambda permission beside InvokeFunction might not be what you want? Not sure!?
W26 Elastic Load Balancer should have access logging enabled
W27 Security Groups found ingress with port range instead of just a single port
W28 Resource found with an explicit name, this disallows updates that require replacement of this resource
W29 Security Groups found egress with port range instead of just a single port
W31 S3 Bucket likely should not have a public read acl
W32 CodeBuild project should specify an EncryptionKey value
W33 EC2 Subnet should not have MapPublicIpOnLaunch set to true
W34 Batch Job Definition Container Properties should not have Privileged set to true
W35 S3 Bucket should have access logging configured
W36 Security group rules without a description obscure their purpose and may lead to bad practices in ensuring they only allow traffic from the ports and sources/destinations required.
W37 EBS Volume should specify a KmsKeyId value
W38 IOT policy should not allow * action
W39 IoT policy should not allow * resource
W40 Security Groups egress with an IpProtocol of -1 found
[...]

PR #502, opened 2021-01-01
#480 creating plaintext file containing current results from cfn_nag_rules command

phelewski:
#480 

Creates a plaintext file containing all of the current rules from the `cfn_nag_rules` command. This list is auto-generated and updated every time the `rake test:all` command is executed during new rule development.

This doesn't create a dynamic or linked list of each individual rule, but it will put a copy of the current rule list in the root of the repo so that users can get a quick glance of the current rules in place without needing to install the gem and run the `cfn_nag_rules` command locally.


swirle13:
Really wishing this got merged. I hate digging through files just to find a list of the cfn_nag rules and definitions.

PR #531, opened 2021-02-19
Fix rule cloudfront minimum protocol version to adjust for new policy

pethers:


pethers:
Similar to https://github.com/stelligent/cfn_nag/pull/496 , but since no known vulnerability exists in Tls1.2 yet I think is's better to keep the rule name.

pethers:
Can anyone please review this. @arothian

arothian (approved):

PR #545, opened 2021-04-14
Remove unneeded homebrew update

arothian:
As discussed in #267 , the gem uses `brew-gem` and the custom brew tap/formula does not need to be updated as part of the pipeline.


arothian:
Old formula cleanup in https://github.com/stelligent/homebrew-tap/pull/7

Issue #568, opened 2021-10-06
SARIF Format Support on Output for cfn_nag


NickLiffen:
Hey 👋 

I didn't know where else to post this so thought I would open an issue. If this is better moved to a GitHub discussion let me know. 

I'm Nick from GitHub, specifically on the GitHub Advanced Security (GHAS) product. I am a big user of `cfn_nag`, and I think it's great to help people track possible security patterns within their cloud formation templates. 

One thing I would **love** is for this tool to support a [SARIF](https://docs.oasis-open.org/sarif/sarif/v2.0/sarif-v2.0.html) output. SARIF is an industry standard for static analysis. 

We have seen many other linting tools support SARIF such as [eslint](https://www.npmjs.com/package/@microsoft/eslint-formatter-sarif), as well as other security tools: [CodeQL](https://codeql.github.com/docs/), etc. 

So, why am I asking for SARIF support? Right now, GitHub Code Scanning (as part of GitHub Advanced Security) supports any file which is in the format of SARIF. That means any data can be uploaded to Code Scanning. We are seeing more and more teams starting to track security and quality alerts within Code Scanning, and I think it would be great if `cfn_nag` results could be uploaded to Code Scanning. 

All I see this being is the following:

```
cfn_nag_scan --input-path template.yml --output-format sarif
```

Right now it supports JSON, etc. If it could also support SARIF, that would be amazing :) 

Love to hear your thoughts 👍  Also happy to discuss this any further 👍 I think this would be a great feature 👍

NickLiffen:
We could then add it as an option on the GitHub Action, which would automatically upload the results to Code Scanning, which I think would be a really nice experience 👍

NickLiffen:
Amazing 💯 Thanks @arothian. I would contribute towards this work but as a developer myself, I must admit Ruby is not my strong suit 😢 

Some tools that may be useful to you:

- [SARIF Validator](https://sarifweb.azurewebsites.net/Validation)
- [SARIF Tutorials](https://github.com/microsoft/sarif-tutorials)

If you would like to validate anything please feel free to let me know 👍

arothian:
> @arothian thanks for this! Is there any way I could run cfn_nag off this branch by any chance? I can just quickly test this by getting the SARIF and uploading it to the GitHub Code Scanning portal.

[...]

PR #573, opened 2021-10-11
SARIF output support

arothian:
Adds support for output results in SARIF format #568 
- Pass rule definitions to result formatters
- Add names to rule definitions/violations
- Utility method to generate violations from rule classes

### DIFF
--- a/cfn-nag.gemspec
+++ b/cfn-nag.gemspec
@@ -5,7 +5,7 @@ require_relative 'lib/cfn-nag/version'
 Gem::Specification.new do |s|
   s.name          = 'cfn-nag'
   s.license       = 'MIT'
-  s.version       = ENV['GEM_VERSION'] || CfnNagVersion::VERSION
+  s.version       = CfnNagVersion::VERSION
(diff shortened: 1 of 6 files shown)


arothian:
Open Questions
- Should the rules list contain all rules in the scan or only those with results?
- Handling of file paths in results

NickLiffen:
It looks like it may be rule specific 🤔 

Screenshot 2021-10-12 at 17 17 20

NickLiffen:
@arothian you are indeed correct, this is using SAM, so they are `AWS::Serverless::Function` set. 

It would be amazing if you could grab the line number before pre-transformed resources. Worst comes to worst (not saying we should do this) we see a lot of companies putting line `1` if they don't know where it is coming from. Most customers find it useful (and they prefer the tool if the actual line numbering works (as that's the value of SARIF and the code scanning GUI). However, if it's too complex architectural (which I understand sometimes these solutions can be), then maybe just default to the line `1`. but only for lines we can't really find.

PR #581, opened 2021-11-03
Update Github Action

arothian:
- Support writing the output to a file
- Add a test for uploading sarif to GH code scanning

### DIFF
--- a/github-action/entrypoint.sh
+++ b/github-action/entrypoint.sh
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+echo "::debug::Using input path: ${INPUT_INPUT_PATH}"
+echo "::debug::Using output path: ${INPUT_OUTPUT_PATH}"
+
+if [ -n "${INPUT_EXTRA_ARGS}" ]; then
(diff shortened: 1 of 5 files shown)


github-action/entrypoint.sh
```diff
@@ -0,0 +1,11 @@
+#!/bin/sh
+
+echo "::debug::Using input path: ${INPUT_INPUT_PATH}"
```
thegonch, line 3: Not sure why the term `INPUT` precedes all of these variable names, especially when you have something like `INPUT_OUTPUT_PATH` that seems to just lend to confusion.  I would suggest we just remove the extra leading `INPUT` for all of these var names.

arothian: Yeah, it's somewhat unpleasant, but GH actions prefixes all configured inputs with `INPUT_` and then exposes them as ENV variables. To the end user of the action though, they would just configure `input_path`, which maps well to the CLI naming. I could change it to `path` but then it would result in that disconnect.

thegonch (approved):

